PR #1405, opened 2014-06-08
Support multiple errors in ValidationToolTipTemplate

PhilPJL:
With INotifyDataErrorInfo in .NET 4.5 it's possible to have more than one validation error per property.
This is fixed easily by modifying the ValidationToolTipTemplate to bind to the errors collection instead of the current item.
I've also modified the popup's FontSize to match the font size of the adorned item. This feels like the correct behaviour to me.

Added some sample code.

One additional (minor) issue that I don't know how to fix, is if a control is inside a viewbox and the validation error is displayed, the popup control doesn't resize while the viewbox is resized. Try it using the sample.


punker76:
added with ba6cf29eeed3079ebd92c6313a4eddf55aeb12c0

PR #1467, opened 2014-07-11
Get the dialog from a MetroWindow

saturne1606:
I need this because i need to fill InputDialog from a barcode scanner.
This function can be used outside from the MahApps.Metro namespace.
This can be modified to not use Reflection, but it should be merged in DialogManager.cs like this
```
public static async Task<TDialog> GetDialogAsync<TDialog>(MetroWindow owner) where TDialog : BaseMetroDialog
{
    TDialog dialog = null;
    await owner.Dispatcher.BeginInvoke(new Action(() =>
    {
        if (owner.metroDialogContainer.Children.Count != 0)
            dialog = owner.metroDialogContainer.Children[0] as TDialog;
    }));
    return dialog;
}
```

Sorry for my bad english...

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/DialogManager.cs
+++ b/MahApps.Metro/Controls/Dialogs/DialogManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Text;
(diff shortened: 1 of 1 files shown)


punker76:
added by 55c1de1d00249f49d1c9b73751e2eed9a081b7c6

PR #1558, opened 2014-09-24
StatusBar style

### LINKED ISSUE #1542: How is Styling of the StatusBar Achieved and Where is it Done?
I am writing an extensible MVVM framework based on Gemini and one of the themes I which to provide is metro-based. My MVVM based Menu and ToolBar theme correctly with MahApps (that is MahApps forms the base style if I do not override and MahApps is included as a `ResourceDictionary`. However, where I am now having a problem is the StatusBar; this does not have an MahApps style (that I can see), so when my framework goes to base the style of the StatusBar on the default one, it is set to null as there is not one. 

How is the StatusBar themed/styled within MahApps metro and where is this done?

punker76:
add a `StatusBar` style to MahApps

2014-09-24_15h58_37

Closes #1542 How is Styling of the StatusBar Achieved and Where is it Done?

### DIFF
--- a/docs/release-notes/1.0.0.md
+++ b/docs/release-notes/1.0.0.md
@@ -5,6 +5,8 @@ but from now on MahApps.Metro follows Semantic Versioning.
 
 # Features
 
+- Added a StatusBar and StatusBarItem style #1558
+
 # Changes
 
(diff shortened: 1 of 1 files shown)


AnuMr:
I just noticed that the color of status bar remains same when the window loses focus, can I control it?

PR #1574, opened 2014-10-07
Update DropDownButton.xaml with MetroCircle-style

netniV:
Added Styling to make the DropDown button look like Metro style.


punker76:
@netniV with #1952 you can set a custom button style via `ButtonStyle` property

PR #1603, opened 2014-10-24
Metro popover support

charlesw:
Support for declaring a popover inspired by Github's and Twitter Bootstrap's popover functionality. Works great with Toggle buttons as per example (checkout other tab in app)!

**Note:** If used inside a tab panel, as used in demo, the tab panel that hosts the target control must have an AdornerDecorator on it otherwise the popover will not work correctly due to a bug\limitation in WPF's implementation (see: http://karlshifflett.wordpress.com/2008/02/19/wpf-validation-errors-disappear-inside-tabcontrol-when-switching-tabitems/).


agc93:
Is this still planned for inclusion? Looks like an awesome alternative to boring old labels!

tempywritescode:
I've put this through it's paces and it works great, I think it should absolutely be included in MA proper.

amkuchta:
Has this got any traction?  I love the though of it and could definitely use it - is it still being worked @tempywritescode @punker76 @charlesw ?!

PR #1669, opened 2014-11-24
Allows the user to disable the window dragging

Alkalinee:
I hate it when the user can move the window with a grid at the window. This property allows me to disable that.


punker76:
@thoemmi the `IsWindowDraggable` can now be changed at runtime 4f407421b9d62e79a04562d877c870bb2ae5cdce

PR #1673, opened 2014-11-25
Add IsWindowDraggable-Property

Alkalinee:
Allows the programmer to disable window dragging from every position. (if there is no titlebar)


punker76:
doublette for #1669

PR #1732, opened 2015-01-06
Expose progressRing Ellipse Diameter Propertie Feature Request

### LINKED ISSUE #1431: Expose Progress Ring Ellipse Diameter Properties
Currently it's set as width/8 which is not flexible for many practical use. It would be nice to expose some of the private properties so users can customize. Thanks.

HugzM:
fixes #1431  
I Change the visibility of EllispeDiameter setter on progressRing control.
It's now possible to set it throughxaml.

the ellipseDiameter is now completly independant from progress ring width.

It's my first contribution to this project and on opensource project in general.
I hope it's match with your expectations

### DIFF
--- a/MahApps.Metro/Controls/ProgressRing.cs
+++ b/MahApps.Metro/Controls/ProgressRing.cs
@@ -23,6 +23,8 @@ namespace MahApps.Metro.Controls
 
         public static readonly DependencyProperty EllipseDiameterProperty = DependencyProperty.Register("EllipseDiameter", typeof(double), typeof(ProgressRing), new PropertyMetadata(default(double)));
 
+		
+
         public static readonly DependencyProperty EllipseOffsetProperty = DependencyProperty.Register("EllipseOffset", typeof(Thickness), typeof(ProgressRing), new PropertyMetadata(default(Thickness)));
 
         private List<Action> _deferredActions = new List<Action>();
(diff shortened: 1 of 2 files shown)


punker76:
@HugzM please make always a branch and after that a pr, thx

thoemmi:
Instead of setting the default value for `EllipseDiameter` in the style in `ProgressRing.xaml`, you should  set it in the declaration of the dependency property, like this:

```C#
public static readonly DependencyProperty EllipseDiameterProperty = 
    DependencyProperty.Register("EllipseDiameter", typeof(double), typeof(ProgressRing), 
        new PropertyMetadata(7));
```

MahApps.Metro/Controls/ProgressRing.cs
```diff
@@ -23,6 +23,8 @@ public class ProgressRing : Control
 
         public static readonly DependencyProperty EllipseDiameterProperty = DependencyProperty.Register("EllipseDiameter", typeof(double), typeof(ProgressRing), new PropertyMetadata(default(double)));
 
+
```
flagbug: Where are these newlines coming from?

MahApps.Metro/Controls/ProgressRing.cs
```diff
@@ -97,8 +99,8 @@ private static void BindableWidthCallback(DependencyObject dependencyObject, Dep
 
             var action = new Action(() =>
             {
-                ring.SetEllipseDiameter(
```
flagbug: Just delete the code, don't comment it out

HugzM:
@punker76 I created a new branch for this request but i can't edit this PR to change the target.

Should I create a new PR ? 
@thoemmi : you're right , i removed the default value and set it by the dependency property

PR #1739, opened 2015-01-07
Pr/1467 Get the dialog from a MetroWindow

punker76:
changes from: Get the dialog from a MetroWindow #1467
/cc @saturne1606


punker76:
added by 55c1de1d00249f49d1c9b73751e2eed9a081b7c6

PR #1748, opened 2015-01-11
Fix ideal foreground for some color styles

Alkalinee:
Here, look at this:
Preview
For the most styles, it looks better with a white title (in my opinion)


flagbug:
Seems like there's an unrelated change to the input dialog in this PR

Alkalinee:
Sorry, i've removed that.

punker76:
thx @Alkalinee

flagbug:
Hm, this is a "breaking change", at least visually, how should we handle this?

robertstefan:
This seems to be happening for a short period of time. The last time I used MA.M [August - September '14] with one of the above accents [Cyan], the text was white. :-/

PR #1750, opened 2015-01-11
InputDialog - select default text

Alkalinee:
If the dialog opens with default text, the text is automatically selected. For example, if the dialog is opened for rename something, it makes it much easier to override all.


punker76:
added with 75604f5f329ef50691512af2599109c8603cd9b2

PR #1756, opened 2015-01-17
Min, Restore/Max animation fix

### LINKED ISSUE #1275: Issues when switching MetroWindow between monitors while maximized
Do this (you need a 2 monitor system):

-create a basic WPF application with a MetroWindow
-run the program, maximize the window on one monitor
-send the window to the other monitor using Shift-Windows key and an arrow key (eg. Shift-Windows-Right Arrow to send from the left monitor to the right)
-try to send it back to the first monitor using the corresponding keyboard shortcut (eg. Shift-Windows-Left Arrow)

For me, this always fails when trying to send from right to left: I press Shift-Windows-left, and the window is still on the right monitor, maximized. However, if I then "Restore Down", the window will jump to the left monitor.

If I start off with the window on the right maximized, then do Ctrl-Windows-left, then Ctrl-Windows-right, and then Restore Down, then window disappears! It's still there in the taskbar, but I can't seem to get it back no matter what I do.

 Both monitors are the same size, however the working area of the right monitor is slightly larger due to the taskbar taking up space on the left monitor. I'm not sure if the difference is because of a resize happening at the same time, or something about switching back to the primary monitor.

### LINKED ISSUE #1539: some unexpected behaviors from minimize-restore-animations changes
i reverted some changes from the pr ##1514 minimize-restore-animations.

what happens:

- if a window starts with ```WindowState = "Maximized"``` the window is not completely maximized
- nasty window painting, if a user maximize or minimize the window (caused by ```WindowStyle = WindowStyle.SingleBorderWindow`` )

2014-09-05_13h16_23
2014-09-05_13h16_40

/cc @Amrykid @flagbug @mrahhal @AdditionalPylons any ideas? if it's possible to fix, then it would be nice to enable/disable this, cause some people are not interessted by this animation (maybe)...

punker76:
this pr should fix the missing minimize and maximize/restore animations.
thx to @AdditionalPylons for the inspiration and work to solve this issue (hopefully).

limitations goes to a window that ignores the taskbar (no animation)

Closes #1539 

this closes #1275 too (Issues when switching MetroWindow between monitors while maximized)

### DIFF
--- a/MahApps.Metro/Behaviours/BorderlessWindowBehavior.cs
+++ b/MahApps.Metro/Behaviours/BorderlessWindowBehavior.cs
@@ -180,27 +180,31 @@ namespace MahApps.Metro.Behaviours
 
         private void HandleMaximize()
         {
-            var metroWindow = AssociatedObject as MetroWindow;
-            var ignoreTaskBar = metroWindow != null && metroWindow.IgnoreTaskbarOnMaximize;
-            if (AssociatedObject.WindowState == WindowState.Maximized && ignoreTaskBar)
+            if (AssociatedObject.WindowState == WindowState.Maximized)
             {
                 // remove resize border and window border, so we can move the window from top monitor position
                 windowChrome.ResizeBorderThickness = new Thickness(0);
                 AssociatedObject.BorderThickness = new Thickness(0);
 
-                // WindowChrome handles the size false if the main monitor is lesser the monitor where the window is maximized
-                // so set the window pos/size twice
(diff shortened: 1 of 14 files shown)


mrahhal:
@punker76 this makes the MetroWindow look like this when maximized on my box:
capture

mrahhal:
However, when testing @AdditionalPylons' branch [here](https://github.com/AdditionalPylons/MahApps.Metro/tree/minimize-restore-animations) everything is working fine.

AdditionalPylons:
I also get that bug. Not executing the WmGetMinMaxInfo() hook in BorderlessWindowBehavior should fix the problem. I'm not quite sure if that hook is even needed anymore, but I don't have a proper setup to test.

PR #1759, opened 2015-01-18
Fix thememanager if resource source case in app.xaml does not match

sickboy:


punker76:
i don't know if this is the real fix for this... the names should be case sensitive.
^ cc @flagbug

flagbug:
The Window filesystem is case insensitive, so we should probably be too I guess

PR #1761, opened 2015-01-19
Don't maximize over autohidden taskbar

AdditionalPylons:
Move AutoHidden taskbar handling code into WindowChromeWorker. 

GetMinMaxInfo hook shouldn't be needed in BorderlessWindowBehavior any longer.

### DIFF
--- a/MahApps.Metro/Microsoft.Windows.Shell/WindowChromeWorker.cs
+++ b/MahApps.Metro/Microsoft.Windows.Shell/WindowChromeWorker.cs
@@ -988,6 +988,17 @@ namespace Microsoft.Windows.Shell
 
             if (force || frameState != _isGlassEnabled)
             {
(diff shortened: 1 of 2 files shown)


MahApps.Metro/Microsoft.Windows.Shell/WindowChromeWorker.cs
```diff
@@ -933,17 +988,6 @@ private void _UpdateFrameState(bool force)
 
             if (force || frameState != _isGlassEnabled)
             {
-                if (SystemParameters.MinimizeAnimation && _chromeInfo.IgnoreTaskbarOnMaximize == false/* && _chromeInfo.UseNoneWindowStyle == false*/)
```
punker76: why removing this? so it's not possible to change `IgnoreTaskbarOnMaximize` at runtime

PR #1763, opened 2015-01-20
Fix the cursor of the Button in ProgressDialog

mrahhal:
Should be "Arrow" instead of inheriting "Wait".


punker76:
thx @mrahhal

PR #1768, opened 2015-01-24
Remove redundant ItemsControl inside our ItemsControl

mrahhal:
We were using an ItemsControl inside our ItemsControl and binding the
nested ItemsControl's source to our ItemsControl's source. Wut?


punker76:
thx @mrahhal

PR #1774, opened 2015-01-28
Fix the chrome for when there is no enough space

### LINKED ISSUE #1773: Chrome panel incorrect behavior when there is no enough space
The chrome panel renders its components in a way that hides the window buttons (the most important component) first when there is no enough space.
capture_old

mrahhal:
This PR will correct the behavior of the chrome bar when there is no enough space to render all components (icon, title, window commands, window buttons...).

Old:
capture_old

The new chrome bar will now handle this correctly:
capture1

This works no matter how much you decide to shrink the window's width, the window buttons are our first priority.

Closes #1773

### DIFF
--- a/MahApps.Metro/Themes/MetroWindow.xaml
+++ b/MahApps.Metro/Themes/MetroWindow.xaml
@@ -32,6 +32,7 @@
                                StrokeThickness="0" />
 
                     <DockPanel LastChildFill="True"
+                               Panel.ZIndex="3"
                                Grid.Row="0"
                                Grid.RowSpan="2"
                                Grid.Column="0">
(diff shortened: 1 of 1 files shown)


mrahhal:
@punker76 I noticed that flyouts work a bit differently now when interacting with the chrome bar. Nothing wrong, just different. Can you check out if the behavior of the flyouts are acceptable like this?

zsKengren:
I tried to apply this fix and the had seen that the flyouts were not working. All of the titlebar elements were on top of the flyout even if the overlay behavior was set to never. 

@punker76  Is this in the plan to get this fixed?

mrahhal:
@zsKengren I knew that the flyouts were not working properly a long time ago. But I think this is the only way to fix #1773, so we'll have to make the flyouts compatible with the new layout.
I'm not sure how the flyouts work so I'll have to spend some time if I want to look into this later.

Issue #1766, opened 2015-01-21
MetroDialog main message control  lost textWrapping


bloodyroll:
the attaching file:
![mahscreen1](https://cloud.githubusercontent.com/assets/5917126/5835472/3fbc4198-a17a-11e4-8f3e-4caa909397cd.png)

soend:
Anybody know how to fix this?

PR #1776, opened 2015-01-29
Text wrapping for MessageDialog Message

soend:
Added back Text wrapping for MessageDialog Message (#1766)


punker76:
@soend the first one is ok, the second change not, the buttons are good so, negative buttons should be different.
/cc @flagbug

PR #1785, opened 2015-02-03
Update readme.md

ButchersBoy:
Add Dragablz as a compatible control (especially as it now has MahApps styles included)


punker76:
:+1:  :+1: @ButchersBoy

PR #1787, opened 2015-02-05
Upgrade WindowPlacementSettings on version change

### LINKED ISSUE #1736: SaveWindowPosition does not upgrade WindowPlacementSettings on version change
When SaveWindowPosition is attached, the WindowPlacementSettings are not "upgraded" when the hosting program's version changes.

I don't see a way for a host to access the WindowPlacementSettings directly because they are protected and marked internal. It would also require a class definition to deserialize the window placement settings for each derived window type (I think). It seems possible the MahApps.Metro Library could upgrade these settings. 

It appears based on my limited understanding that this could be accomplished as follows:

- Add a `UpgradeSettings` property to `IWindowPlacementSettings`, which defaults to `true`
- Add a `void Upgrade()` method to `IWindowPlacementSettings`, which is implemeneted by `ApplicationSettingsBase`
- If on loading the `UpgradeSettings` flag is `true`, call `_settings.Upgrade()`, set the flag to `false`, and then `_settings.Save()`

This is the recommended way user settings are handled by programs.

punker76:
- Added a `UpgradeSettings` property to `IWindowPlacementSettings`, which defaults to true
- Added a `void Upgrade()` method to `IWindowPlacementSettings`, which is implemeneted by `ApplicationSettingsBase`

Closes #1736

### DIFF
--- a/MahApps.Metro/Controls/WindowSettings.cs
+++ b/MahApps.Metro/Controls/WindowSettings.cs
@@ -2,7 +2,6 @@
 using System.ComponentModel;
 using System.Configuration;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Interop;
@@ -18,11 +17,6 @@ namespace MahApps.Metro.Controls
         /// Refreshes the application settings property values from persistent storage.
         /// </summary>
         void Reload();
-
-        /// <summary>
-        /// Restores the persisted application settings values to their corresponding default properties.
-        /// </summary>
-        void Reset();
 
         /// <summary>
         /// Upgrades the application settings on loading.
@@ -100,14 +94,6 @@ namespace MahApps.Metro.Controls
         protected virtual void LoadWindowState()
         {
             if (_settings == null) return;
-
-            if (_settings.UpgradeSettings)
-            {
-                _settings.Upgrade();
-                _settings.UpgradeSettings = false;
-                _settings.Save();
-            }
-
             _settings.Reload();
 
             // check for existing placement and prevent empty bounds
--- a/MahApps.Metro/Controls/WindowSettings.cs
+++ b/MahApps.Metro/Controls/WindowSettings.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Configuration;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Interop;
@@ -12,7 +13,31 @@ namespace MahApps.Metro.Controls
     public interface IWindowPlacementSettings
     {
         WINDOWPLACEMENT? Placement { get; set; }
+
+        /// <summary>
+        /// Refreshes the application settings property values from persistent storage.
+        /// </summary>
         void Reload();
+
+        /// <summary>
+        /// Restores the persisted application settings values to their corresponding default properties.
+        /// </summary>
+        void Reset();
+
+        /// <summary>
+        /// Upgrades the application settings on loading.
+        /// </summary>
+        [DefaultValue(true)]
+        bool UpgradeSettings { get; set; }
+
+        /// <summary>
+        /// Updates application settings to reflect a more recent installation of the application.
+        /// </summary>
+        void Upgrade();
+
+        /// <summary>
+        /// Stores the current values of the settings properties.
+        /// </summary>
         void Save();
     }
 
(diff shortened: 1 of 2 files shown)


punker76:
@mike-ward is this what you need? is this enough?

mike-ward:
I think the `Upgrade` method is all that is really needed. It's typical for apps to have their own `UpgradeSettings` property.

punker76:
@mike-ward so the others (at th einterface) too much?

mike-ward:
Probably. I can't think of a use case where I would want to reset or reload those settings. For my use case, 'Upgrade` is sufficent.

punker76:
@mike-ward ok, you're right. i removed the reset method, but leave the save method, i think that's also recommended.

flagbug:
Hm, that's a breaking change, but I guess it's alright

mike-ward:
Upgraded to 1.1.0. Could I get a quick tutorial on how to upgrade settings? Here is what I've tried.

In App.xaml.cs, `OnStartup()`

    var windowPlacementSettings = mainWindow.GetWindowPlacementSettings();

    if (windowPlacementSettings.UpgradeSettings)
    {
        windowPlacementSettings.Upgrade();
        windowPlacementSettings.UpgradeSettings = false;
        windowPlacementSettings.Save();
    }

UpgradeSettings is always false. I must be missing something else where?

punker76:
it's a mistake with `DefaultValue(true)` and auto property, i'll fix this
tomorrow and publish a 1.1.1 version
Am 25.02.2015 00:56 schrieb "Mike Ward" <notifications@github.com>:

mike-ward:
Cool beans! The window snap restore thing is working nicely.

punker76:
@mike-ward `UpgradeSettings` should now work :-D 509bd7743e0e31fc3f63cd419432d633869ad743

mike-ward:
Different problem. Using code from above, `UpgradeSettings` is always `true` now.

punker76:
@mike-ward to clarify this, i will change the implementation to this

`` csharp
        /// <summary>
        /// Upgrades the application settings on loading.
        /// </summary>
        [UserScopedSetting]
        public bool UpgradeSettings
        {
            get
            {
                if (this["UpgradeSettings"] != null)
                {
                    return (bool)this["UpgradeSettings"];
                }
                return true;
            }
            set { this["UpgradeSettings"] = value; }
        }
```

with the result

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
    <configSections>
        <sectionGroup name="userSettings" type="System.Configuration.UserSettingsGroup, System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" >
            <section name="MahApps.Metro.Controls.WindowApplicationSettings.MetroDemo.ExampleWindows.FlyoutDemo" type="System.Configuration.ClientSettingsSection, System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" allowExeDefinition="MachineToLocalUser" requirePermission="false" />
        </sectionGroup>
    </configSections>
    <userSettings>
        <MahApps.Metro.Controls.WindowApplicationSettings.MetroDemo.ExampleWindows.FlyoutDemo>
            <setting name="UpgradeSettings" serializeAs="String">
                <value>False</value>
            </setting>
            <setting name="Placement" serializeAs="Xml">
                <value>
                    <WINDOWPLACEMENT xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                        xmlns:xsd="http://www.w3.org/2001/XMLSchema">
                        <length>44</length>
                        <flags>0</flags>
                        <showCmd>1</showCmd>
                        <minPosition>
                            <X>-1</X>
                            <Y>-1</Y>
                        </minPosition>
                        <maxPosition>
                            <X>-1</X>
                            <Y>-1</Y>
                        </maxPosition>
                        <normalPosition>
                            <left>1028</left>
                            <top>418</top>
                            <right>1903</right>
                            <bottom>1043</bottom>
                        </normalPosition>
                    </WINDOWPLACEMENT>
                </value>
            </setting>
        </MahApps.Metro.Controls.WindowApplicationSettings.MetroDemo.ExampleWindows.FlyoutDemo>
    </userSettings>
</configuration>
```

is this what you want?

mike-ward:
Yep, I think that will work. Thanks for doing this.

punker76:
@mike-ward np, but can you give me a reason why a user want change this property in this file? i don't see the benefit, maybe i'm blind ;-)

mike-ward:
You only want to upgrade the settings once per version bump. The suggested
way to do this is:

http://blog.johnsworkshop.net/automatically-upgrading-user-settings-after-an-application-version-change/

It doesn't matter to me if I keep a flag in my user settings or use the
MahApp supplied flag. If you decide not to persist the flag then I would
suggest removing it.

PR #1793, opened 2015-02-10
fix preservetext for buttons in dialogs

### LINKED ISSUE #1755: MessageDialog NegativeButton styling
MessageDialog NegativeButton styling is not set to AccentedDialogSquareButton. Would be nice if it would use also this style so text casing would stay same as it is when i set it.

punker76:
there is a little bug with accented dialogs and the text of the buttons, so this should fix it...

Closes #1755


flagbug:
I don't think we need a new property here, the casing should actually *always*  be preserved, but there seems to be a bug where this doesn't happen if the `ColorScheme` is set to  `MetroDialogColorScheme.Accented`

punker76:
@flagbug agreed, changed right now

flagbug:
@punker76 :+1:

PR #1812, opened 2015-02-21
Reversed searching for the theme

Alkalinee:
I had some problems with the theme of the dialogs (I apply the themes by my own) - the dialogs were always in the light style. That was because the merged resources of my application contains two themes - the last theme is the applied, so it doesn't make sense to begin searching for the current theme from zero.

### DIFF
--- a/MahApps.Metro/Styles/Controls.PasswordBox.xaml
+++ b/MahApps.Metro/Styles/Controls.PasswordBox.xaml
@@ -99,14 +99,11 @@
                         </Storyboard>
                     </ControlTemplate.Resources>
                     <Grid>
-                        <Rectangle x:Name="Base"
-                                   Stroke="{TemplateBinding BorderBrush}"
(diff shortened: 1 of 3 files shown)


flagbug:
Hm, I'm not sure if I understand the use case here, do you think you can explain it step-by-step?

Alkalinee:
Yes, okay:
The merged resources of my application look like that:
0. /Resources/Themes/BaseLight.xaml
1. /Resources/Languages/Hurricane.de-de.xaml
2. /Resources/Themes/BaseDark.xaml

The enumerator would detect a light theme, but the application is completely dark because what you see starts at the last resource, but the enumerator starts at the first resource with searching. (if you add a resource dictionary which contains already existing keys, they override the existing)

punker76:
@Alkalinee if you want to replace a theme (or accent) you should use the `ThemeManager` instead placing multiple themes in a Window or App

Alkalinee:
Okay, but I think it would be better if it would start to search from the end because it should detect always the current theme and not the first theme.

PR #1819, opened 2015-02-26
Added alignment properties for MetroDataGridRowHeader

chuuddo:
Now we can use style like this
```csharp
<Style x:Key="RowHeaderStyle" TargetType="DataGridRowHeader"
       BasedOn="{StaticResource MetroDataGridRowHeader}">
    <Setter Property="HorizontalContentAlignment" Value="Center" />
</Style>
```
```csharp
<DataGrid RowHeaderStyle="{StaticResource RowHeaderStyle}" />
```
screenshot_1

### DIFF
--- a/MahApps.Metro/Styles/Controls.DataGrid.xaml
+++ b/MahApps.Metro/Styles/Controls.DataGrid.xaml
@@ -250,6 +250,10 @@
 
     <Style x:Key="MetroDataGridRowHeader"
            TargetType="{x:Type DataGridRowHeader}">
+        <Setter Property="HorizontalContentAlignment"
+                Value="Left" />
+        <Setter Property="VerticalContentAlignment"
+                Value="Center" />
         <Setter Property="Background"
                 Value="Transparent" />
         <Setter Property="BorderBrush"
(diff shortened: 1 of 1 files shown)


MahApps.Metro/Styles/Controls.DataGrid.xaml
```diff
@@ -250,6 +250,10 @@
 
     <Style x:Key="MetroDataGridRowHeader"
            TargetType="{x:Type DataGridRowHeader}">
+        <Setter Property="HorizontalContentAlignment"
+                Value="Left" />
```
punker76: the default should be Stretch

chuuddo: @punker76  dont know, but in MetroDataGridColumnHeader default is Left.
```
<Style x:Key="MetroDataGridColumnHeader"
           TargetType="{x:Type DataGridColumnHeader}">
       ...
        <Setter Property="HorizontalContentAlignment"
                Value="Left" />
        <Setter Property="VerticalContentAlignment"
                Value="Center" />
```

punker76: yeah i know, but the default before your changes is Stretch (it's the default if you doesn't say HorizontalAlignement="xyz")

chuuddo: you are right

PR #1821, opened 2015-02-26
Move window when flyout is open

hercmaroudas:
When a flyout was open you weren't able to move the window where the flyout was over the title bar. This was particularly anoying if you wished to move the window while a flyout was opened as a modal flyout or covered the entire area of the window.

I've implemented a change that enables the window to be moved when a flyout is opened and the mouse cursor is within the windows title bar area. 

If the default behaviour was intended by design perhaps this feature could be optional with a property. 

I hope this is ok.

### DIFF
--- a/MahApps.Metro/Controls/Flyout.cs
+++ b/MahApps.Metro/Controls/Flyout.cs
@@ -487,7 +487,7 @@ namespace MahApps.Metro.Controls
 
             ApplyAnimation(Position, AnimateOpacity);
 
-            SetWindowEvents();
+            SetParentWindowEvents();
         }
 
         internal void ApplyAnimation(Position position, bool animateOpacity, bool resetShowFrame = true)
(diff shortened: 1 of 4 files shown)


MahApps.Metro/Controls/Flyout.cs
```diff
@@ -542,6 +547,24 @@ internal void ApplyAnimation(Position position, bool animateOpacity, bool resetS
...
+            if (window != null)
+            {
+                this.ClearParentWindowEvents(window);
+
+                PART_WindowTitle.MouseDown += window.TitleBarMouseDown;
+                PART_WindowTitle.MouseUp += window.TitleBarMouseUp;
+            }
+        }
+
+        private void ClearParentWindowEvents(MetroWindow window)
```
punker76: this should be also called in `FlyoutsControl` to avoid memory leaks (cause it's possible to have dynamic flyouts)

```csharp
        protected override void ClearContainerForItemOverride(DependencyObject element, object item)
        {
            var window = this.TryFindParent<MetroWindow>();
            if (window != null)
            {
                ((Flyout)element).ClearParentWindowEvents(window);
            }
            base.ClearContainerForItemOverride(element, item);
        }
```

hercmaroudas: Ok thanks Punker, added and sent the pull request.

punker76:
@hercmaroudas looks good, i'll be add this to 1.2.0

hercmaroudas:
Thanks @punker76 :+1:

thoemmi:
Looks ok, though I don't like that `Flyout` accesses `MetroWindow` directly, but I have no better idea. 

However, I would suggest some changes to your *Flyout.xaml*: In your PR you move the flyout's content down by the height of the titlebar. I'd prefer to keep the content where it is but **overlay** it with the `PART_WindowTitle`. So I'd remove the `RowDefinitions` again, add the `Rectangle` **after** the `AdornerDecorator`, and set its `VerticalAlignment` to `Top`:
```
    <AdornerDecorator>
        <!-- -->
    </AdornerDecorator>
    <Rectangle x:Name="PART_WindowTitle"
               Height="{Binding TitlebarHeight, Mode=OneWay, RelativeSource={RelativeSource FindAncestor, AncestorLevel=1, AncestorType={x:Type Controls:MetroWindow}}}"
               HorizontalAlignment="Stretch" VerticalAlignment="Top"
               StrokeThickness="0"
               Fill="Transparent"/>
```

PR #1822, opened 2015-02-27
fix Gripper for MetroDataGridRowHeader

chuuddo:
Now grippers displayed in rowheaders.
image

### DIFF
--- a/MahApps.Metro/Styles/Controls.DataGrid.xaml
+++ b/MahApps.Metro/Styles/Controls.DataGrid.xaml
@@ -94,18 +94,21 @@
 
     <Style x:Key="MetroRowHeaderGripperStyle"
            TargetType="{x:Type Thumb}">
-        <Setter Property="Width"
-                Value="8" />
+        <Setter Property="Height"
+                Value="6" />
         <Setter Property="Background"
-                Value="Transparent" />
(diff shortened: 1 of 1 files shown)


MahApps.Metro/Styles/Controls.DataGrid.xaml
```diff
@@ -94,18 +94,21 @@
 
     <Style x:Key="MetroRowHeaderGripperStyle"
            TargetType="{x:Type Thumb}">
-        <Setter Property="Width"
-                Value="8" />
+        <Setter Property="Height"
+                Value="6" />
```
punker76: i think this should be
```xaml
        <Setter Property="Width"
                Value="8" />
        <Setter Property="Height"
                Value="6" />
```

chuuddo: i dont think so
image

punker76: ok, but the you should replace the `Rectangle` with the `Border` again, cause `Rectangle` must specify a width

punker76: ok, you're right, my mistake

PR #1825, opened 2015-03-02
[WIP] Header, headers, headers

flagbug:
The Windows 8/WinRT/ModernUI/Whatever XAML framework has the really cool feature of "headers" for some controls

These are two `TextBoxes`, declared in XAML as following;

``` xml
<StackPanel Orientation="Vertical">
    <TextBox Header="Username" Height="50" Width="150" />
    <TextBox Header="Password" Height="50" Width="150" />
</StackPanel>
```

This PR adds this feature via an attached property to some of the controls

**TODO**
- [ ] How big should the header be?
- [ ] Implement it for all TextBoxes
- [ ] `HeaderTemplate` property because people want to customize
- [ ] Additional controls?

### DIFF
--- a/MahApps.Metro/Controls/Helper/ControlsHelper.cs
+++ b/MahApps.Metro/Controls/Helper/ControlsHelper.cs
@@ -10,6 +10,20 @@ namespace MahApps.Metro.Controls
     /// </summary>
     public static class ControlsHelper
     {
+        public static readonly DependencyProperty HeaderProperty =
+            DependencyProperty.RegisterAttached("Header", typeof(object), typeof(ControlsHelper), new FrameworkPropertyMetadata(null));
+
+        [AttachedPropertyBrowsableForType(typeof(TextBox))]
+        public static object GetHeader(UIElement element)
+        {
+            return (object)element.GetValue(HeaderProperty);
+        }
+
+        public static void SetHeader(UIElement element, object value)
+        {
+            element.SetValue(HeaderProperty, value);
+        }
+
         public static readonly DependencyProperty HeaderFontSizeProperty =
             DependencyProperty.RegisterAttached("HeaderFontSize", typeof(double), typeof(ControlsHelper), new FrameworkPropertyMetadata(26.67, HeaderFontSizePropertyChangedCallback){ Inherits = true});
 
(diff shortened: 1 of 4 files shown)


thoemmi:
It's very similar to the existing `TextBoxHelper.UseFloatingWatermark`, but I like it :smile: 
image

punker76:
:octopus: :+1:

jimbobsmith:
Just my 2 cents.  The way I handled this in my projects was to create a header control which acts as a container for any input control.  In this way, I only had to create a single control and can easily use it for any type of input control.  I used a Label internally and handled the target binding internally etc.  It is a bit more work to use that than to just add a property value but also avoids having to care about input control specific implementations.  
e.g.
`<Header Header="Some _Header>`
`  <TextBox Text={Binding SomeTextValue} />`
`</Header>`

Amrykid:
:+1:

amkuchta:
@punker76 @flagbug @thoemmi @jimbobsmith @Amrykid , I know that this is an old item, but I cam across it and had a thought about it - I'm just not sure how well it will be received...  

What if, instead of having the `Header` be a separate property, we somehow modified the location of the floating watermark (e.g `TextBoxhelper.WatermarkFloatLocation="Interior"` or `TextBoxHelper.WatermarkFloatLocation="Exterior"`).  You could even go a step farther by picking side (e.g. `InteriorTop`, `InteriorLeft`, `ExteriorBottom`, `ExteriorRight`, etc.).  Again, I am not sure as to how well this merges the two ideas, but it might be a nifty way to utilize a property that already exists.

Thoughts...?

punker76:
@amkuchta I don't like the idea to enhance the watermark. I'm more with a new control like the Badged control. A ContentControl which can take any stuff and surround it with a Header text...

thoemmi:
I agree with @punker76. Watermarks and headers are two different paradigms which shouldn't be mixed.

amkuchta:
@punker76 @thoemmi sounds good to me - I still would like to see the external floating watermark in addition to the new `Header` control (I don't want headers that always show, only when there is text in the control.  When the control is empty, I want the instructions to be a watermark simply because I think it looks cleaner that way).  I generated some code and _accidentally_ rolled it into #2892 :rofl: .  However, I am now stuck on it :hankey: .  Any insight/help you guys could provide would be awesome.  I'm trying to become a more active contributor instead of being a sponge who pushes ideas and issues but never offers to help...

PR #1833, opened 2015-03-11
Add tooltip text converter to RangeSlider

Alkalinee:
The changes allow to use a custom format for the tooltip. For example, I
want that the user can choose a timespan (ms). The tooltip just shows
large numbers. With a converter, the tooltip can display the value in a
time format (mm:ss) which is much better to read for the user.
Preview


QuantumDeveloper:
cool!

punker76:
@Alkalinee thx

PR #1852, opened 2015-03-25
Metro touchable thumb of Slider

ycaoyang:
In touch environment, touching a slider won't make its thumb following user's finger, for another visual has capture the touch-device. In order to make the Thumb touchable, we have to capture the touch-device in himself. So here it is.


flagbug:
I think you selected the wrong branch for your PR

ycaoyang:
So do I. Sorry for the mistake, and thanks for reply. I'll push a new one after a while.

PR #1853, opened 2015-03-25
create new class of Thumb in order to be better in Touch environment

ycaoyang:
In touch environment, touching a slider won't make its thumb following user's finger, for another visual has capture the touch-device. In order to make the Thumb touchable, we have to capture the touch-device in himself. So here it is.

PS : correct branch selection

### DIFF
--- a/MahApps.Metro/Controls/MetroTouchThumb.cs
+++ b/MahApps.Metro/Controls/MetroTouchThumb.cs
@@ -1,0 +1,53 @@
+﻿using System.Windows.Controls.Primitives;
+using System.Windows.Input;
+
+namespace MahApps.Metro.Controls
+{
+    public class MetroTouchThumb : Thumb
+    {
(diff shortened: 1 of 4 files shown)


Amrykid:
@ycaoyang Make sure to add your thumb class to the same directory in the MahApps.Metro.NET45 solution as well.

punker76:
@All can someone test this? I don't have a touch device...

Kalashn:
is this fix present in one of the current alpha releases available on nuget? if so, i'll test it with pleasure (on only one touch device though)

punker76:
@Kalashn you can clone this pr `https://github.com/ycaoyang/MahApps.Metro/tree/y.cao`

Kalashn:
Thanks, I quite new to github. I'll run the tests this weekend. Any procedures or tools I have to use? Otherwise it will be just a basic: is it working as expected (and proposed as better) test case.

Issue #1850, opened 2015-03-24
CustomDialog in Visual Studio Designer throws exception


marcpiechura:
HI, 

I would like to see the preview of my CustomDialog in Visual Studio but unfortunately it shows the folling error message.

![image](https://cloud.githubusercontent.com/assets/11612539/6802297/82c9439c-d22f-11e4-8755-855c3cb2c4e4.png)

The fix is realy simple, in BaseMetroDialog -> HandleTheme add the NULL check for windowTheme

```

                var windowTheme = DetectTheme(this);

	            if (windowTheme == null)
		            return;


                var theme = windowTheme.Item1; <-- Throws
```

Thanks!

johnsimmons:
I apologize if this is the incorrect place to post a question, but I didn't want another duplicate of this problem (I'll remove it and ask somewhere else if needed).  

I'm having this same problem with the newest version of MahApps.Metro.  The mentioned code fix is in place but my problem is identical to the original post by marcpiechura.

I added a ```UserControl (WPF)``` and added this code to my new xaml file

```<Dialogs:CustomDialog x:Class="HR_DesktopApplication.MyCustomDialog"
             xmlns:Dialogs="clr-namespace:MahApps.Metro.Controls.Dialogs;assembly=MahApps.Metro"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
             mc:Ignorable="d" 
             d:DesignHeight="300" d:DesignWidth="300">
    <Grid>
        
    </Grid>
</Dialogs:CustomDialog>```

I imagine this is human error (since I'm new to wpf in general), but I wanted to make sure it wasn't the same bug.

marcpiechura:
@johnsimmons 
You need to upgrade to the latest 1.1.3 alpha, it's not included in 1.1.2. 

@punker76 
Maybe you could change the milestone to 1.1.3 for clarification ?

Regards

johnsimmons:
Awesome, thanks for the response.  I'm glad it was just a misunderstanding on my part.

PR #1857, opened 2015-03-31
Added windowTheme NULL check.

marcpiechura:
See issue #1850

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
@@ -6,347 +6,351 @@ using System.Windows.Media.Animation;
 
 namespace MahApps.Metro.Controls.Dialogs
 {
-    /// <summary>
-    /// The base class for dialogs.
-    ///
-    /// You probably don't want to use this class, if you want to add arbitrary content to your dialog, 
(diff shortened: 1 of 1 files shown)


Amrykid:
Looks good. :+1: 

@flagbug @punker76 ?

punker76:
@Amrykid we can do this, but what i see is, that this is in some cases (forgot to add a custom theme or accent) a user failure... so maybe we should handle this with our own exception...

thoemmi:
If I understand the original issue #1850 correctly, this happens only in the Visual Studio designer. Maybe you should do the check for `null` only at design time (hint: [DesignerProperties.GetIsInDesignMode](https://msdn.microsoft.com/en-us/library/system.componentmodel.designerproperties.getisindesignmode.aspx))

marcpiechura:
@thoemmi that's right, if I run the application everything is fine.

punker76:
@thoemmi @marcpiechura so let's check for designer mode...

marcpiechura:
@punker76 alright, should I update the pull request or do you want to add it yourself?

punker76:
@marcpiechura simply update this pr

punker76:
thx @marcpiechura i added your changes manually in 3f601ab

PR #1862, opened 2015-04-03
DataGrid ContextMenu for Column Visibility

mouadhtr:
Add a context Menu to data grid control columns to
Hide or show the columns by right-clicking
on any of the data grid’s column headers.


punker76:
I think that's too specific for MahApps

PR #1863, opened 2015-04-03
Update FlyoutDemo.xaml

smalls89:
Found a closing tab for exampleWindows:CustomFlyout that hasn't an opening tag, guessing it was left over from something... Also, demo runs fine without it.


smalls89:
i lied, it no runs

punker76:
there are 12 flyouts

the custom flyout is

```
        private void ShowThird(object sender, RoutedEventArgs e)
        {
            this.ToggleFlyout(2);
        }
```

smalls89:
looked at it again... it has been a long day...

PR #1892, opened 2015-04-19
The Margin of ContentPresenter in MetroListBoxItem

ye4241:
The Margin of ContentPresenter in MetroListBoxItem should inherit {TemplateBinding Padding}, and let developer use the Padding the change the Margin of ContentPresenter


punker76:
@ye4241 thx

ye4241:
@punker76 :smirk:

PR #1902, opened 2015-04-26
WindowCommands layout fix

Dotnetifier:
Noticed a tiny annoyance: the WindowCommands' height did not span the entire height of the title bar, which makes the buttons harder to reach (which is my opinion as I just throw the cursor to the top of the screen to click). I'm sure this behavior used be there a while ago.

From this
inactive
to this
active


punker76:
changed by ee14732fe83810583ecf7e6027b8c8d87849b8b4

Issue #1910, opened 2015-04-30
awaiting ShowMetroDialogAsync  has no effect


IsNull:
_Using the latest nuget package of mahapps (1.1.2.0), .NET 4.5_

I wonder why custom dialoges behave differently than the metro message box, in that the custom dialoges Task does not wait for the dialoge to close. I have found similar questions & bug reports but none of them have a acceptable answer, such as #1227.

```csharp
var controller = await _metroWindow.ShowMessageAsync( ... );
// As expected, awaits until the user has choosen a button
```
```csharp
await _metroWindow.ShowMetroDialogAsync( ... );
// await has no effect at all - behaves like a fire and forget async call
```

* Is this by design? Then why is there a task returned on ShowMetroDialogAsync ?
* Is this a bug? I may be able to look into this and do a pull request.
* Or is there another way to await the user response from this dialog asynchronously.

Can someone please share some thoughts behind this API?

flagbug:
Yes, this is by the design, it's noted in the method summary: https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Controls/Dialogs/DialogManager.cs#L292

IsNull:
If I see that correctly in the code, you did not implement awaiting since you had the problem that "dialog.WaitForButtonPressAsync()" is not available in the custom dialog?

Maybe it can be solved generically by hooking into the closing of the dialog, i.e. instead of waiting for a button pressed, wait until the dialog is gone...

I am happy to give it a shot and implement this bit, so the returned task actually awaits the closing of the dialog. Would you mind a pull request? Or do you have other plans with this api?

IsNull:
After some testing, I finally have it working now in my fork https://github.com/ElderByte-/MahApps.Metro/commit/62557596a266137b5d19b067afd61560a110dd89

I added a simple method to the BaseMetroDialog
```csharp
        public Task WaitUnitlClosed()
        {
            TaskCompletionSource<object> tcs = new TaskCompletionSource<object>();

            Unloaded += (s,e) =>
            {
                tcs.TrySetResult(null);
            };

            return tcs.Task;
        }
```

With this, I added a new Method to the DialogManager (optionally we can also replace the old Async Dialog Method)

```csharp
        /// <summary>
        /// Adds a Metro Dialog instance to the specified window and makes it visible.
        /// You can await until this dialog is removed/hidden by awaiting the returned task.
[...]

PR #1911, opened 2015-04-30
ShowMetroDialogAsync - Dialog is awaitable

IsNull:
As discussed in issue #1910, the ShowMetroDialogAsync method has now the same behaviour as the ShowMessageAsync has. If you await the returned task, it will wait until the dialog is gone.

This pull request contains the necessary changes and adds also a demo of this new capbability to the show case.

The now working code in the show-case pretty much sais all:

```csharp
await this.ShowMetroDialogAsync(dialog);

await dialog.WaitUntilUnloadedAsync(); // Just call await here to wait for the dialog to go away

// This message box is displayed after the dialog has been closed by the user
await this.ShowMessageAsync("Dialog gone", "The custom dialog has closed");
```

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
@@ -253,7 +253,7 @@ namespace MahApps.Metro.Controls.Dialogs
         /// Waits until this dialog gets unloaded.
         /// </summary>
         /// <returns></returns>
-        public Task WaitUnitlUnloaded()
+        public Task WaitUntilUnloadedAsync()
         {
             TaskCompletionSource<object> tcs = new TaskCompletionSource<object>();
 
--- a/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
@@ -249,8 +249,11 @@ namespace MahApps.Metro.Controls.Dialogs
         /// </summary>
         protected internal MetroWindow OwningWindow { get; internal set; }
 
-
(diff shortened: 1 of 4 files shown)


flagbug:
This change may break the code of other people

IsNull:
> This change may break the code of other people

Agreed. I will refactor it tomorrow so we have a new method and the old one will behave as it had.
Maybe we can mark it as obsolete and remove it in the next major release.

IsNull:
I have now created a 2nd method which has the awaitable feature, for now it is called _ShowMetroDialogAsync**Awaitable**_ so existing code should not break with this change.

If you come up with a better name, I am happy to change it.

IsNull:
I have once more iterated and made the changes way simpler. 
The exisiting API is unchanged, but if you want to wait until the dialog is gone, just call 
```csharp
await dialog.WaitUnitlUnloaded();
```
on your custom dialog.

MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
```diff
@@ -249,6 +249,23 @@ internal protected virtual bool OnRequestClose()
         /// </summary>
         protected internal MetroWindow OwningWindow { get; internal set; }
 
+        /// <summary>
+        /// Waits until this dialog gets unloaded.
+        /// </summary>
+        /// <returns></returns>
+        public Task WaitUnitlUnloaded()
```
flagbug: Typo here, should be `WaitUntilUnloaded`

IsNull:
Fixed the typo and also renamed the method to `WaitUntilUnloadedAsync` to be in line with the other task returning methods.

punker76:
thx @IsNull

PR #1943, opened 2015-05-29
add mechanism for easy access of a dialog from a view model

ButchersBoy:
Using this mechanism you can easily use the MahApps dialogs from a view model.  It is agbostic to whatever DI framework you may or may not be using.

To use you add an attached property in your XAML, which will tie the Binding (your view model) to the Window:

```
Dialog:DialogParticipation.Register="{Binding}"
```

Then the DialogCoordinator can be called from your view model, providing it with a context (your view model), it will know which Window to use to show the dialog.

```
_dialogCoordinator.ShowInput(this).ContinueWith(t => Console.WriteLine(t.Result));
```

__outstanding:__ I've added to the demo app, but only a single dialog type right now.  If you accept this PR I will build out the interface for the other dialog types.

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/DialogCoordinator.cs
+++ b/MahApps.Metro/Controls/Dialogs/DialogCoordinator.cs
@@ -11,7 +11,58 @@ namespace MahApps.Metro.Controls.Dialogs
         /// </summary>
         public static readonly DialogCoordinator Instance = new DialogCoordinator();
 
-        public Task<string> ShowInput(object context)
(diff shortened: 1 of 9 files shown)


ButchersBoy:
image

flagbug:
This will totally lock us in to a fixed amount of different dialogs.

 Because of API versioning, we now can't add for example a `ShowCoolNewDialogAsync`.

punker76:
@flagbug i think there is no need to implement more special dialogs, cause we have cusrom dialog...

ButchersBoy:
Yeah, you have custom dialogs, and also, the new interface is only a facade onto the existing methods on the window...don't see how that would cause lock in.

PR #1947, opened 2015-05-31
Setting for default focused button in MessageDialog

PeterHageus:
Introduces new setting in MetroDialogSettings, to determine which button get focus by default in MetroDialog. Need this so user can hit enter when Affirmative is the default option.


BlythMeister:
Hi,

I'm really needing this feature, but can see there was a merge conflict.

I have opened a new PR #2265 to replace this one (i couldn't push into @PeterHageus fork)

PR #1951, opened 2015-06-02
1950 some classes crashe if window is closed immediately after showing it

szogun1987:
Mine proposition of bug fix


punker76:
-> 554cef8aee166b95c3bb7daf0e0de9006c646bcc

PR #1958, opened 2015-06-05
Watermark for non editable combobox

### LINKED ISSUE #1953: Watermark on ComboBox where IsEditable="False"
Currently, the only way to display a watermark on a ```ComboBox``` control is to set ```IsEditable="True"```.  Is there any way to display a watermark when this is set to ```False```?  I want to have prompt text for the user but cannot give them the option to type because of a converter that I have in place.

Thanks in advance!

punker76:
watermark (not editable and not focused)
image

watermark (focused)
image

Closes #1953 Watermark on ComboBox where IsEditable="False"


amkuchta:
@punker76 this is perfect, thank you!

hunyeti:
Can this be used before the release of a new mahapps version?

PR #1962, opened 2015-06-08
Possiblity to define the UpDownButtons width

Prophetebe:
Add the possibility to define the UpDownButtons width in
DataGridNumericUpDownColumn. This can be usefull for a touch application
because by default, a width of 20 is not enough because the buttons are
too close


punker76:
@Prophetebe thx, that makes sense, but can you look to my comments?

punker76:
@Prophetebe thx, i merged this with a little change to the column definitions width

Prophetebe:
Thank you. Happy to help for this great tools :).

PR #1979, opened 2015-06-13
RecognizesAccessKey in styles for CheckBox, RadioButton, and GroupBox

thoemmi:
The styles for CheckBox, RadioButton, and GroupBox are missing `RecognizesAccessKey="true"`, so keyboard navigation fails when using shortcuts (`_` in front of access keys)


punker76:
@thoemmi :+1:  thx

PR #1988, opened 2015-06-17
Expander enhancements

punker76:
- [x] some styling fixes (margin, padding, border thickness)
- [x] add `ExpanderHelper` with
  + 4 new attached dependency properties `HeaderUpStyle`, `HeaderDownStyle` and `HeaderLeftStyle`, `HeaderRightStyle`. so it's now possible to create your own header toggle styles without rewriting the `Expander` style
- [x] better handling for `BorderThickness`
  + the `Expander` uses now the new value converter `ThicknessBindingConverter`, so setting the `BorderThickness` is much better than before
- [x] Click on Expander header support #1291 (just do the default expander behavior)
- [x] new `PreserveTextCase` attached property

and for `GroupBox`
- [x] some styling fixes (margin, padding, border thickness)
- [x] `PreserveTextCase` attached property
- [x] `ThicknessBindingConverter`, so setting the `BorderThickness` is much better than before

image

image

image

image

### DIFF
--- a/MahApps.Metro/Controls/Helper/ControlsHelper.cs
+++ b/MahApps.Metro/Controls/Helper/ControlsHelper.cs
@@ -10,7 +10,7 @@ namespace MahApps.Metro.Controls
     /// </summary>
     public static class ControlsHelper
     {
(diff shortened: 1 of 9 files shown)


punker76:
^ @xxMUROxx Dotnetifier it's time for the click on header behavior :-D

PR #1997, opened 2015-06-21
creating a Task instance doesn't execute it

thoemmi:
Because of .NET 4.0 compatibility, MahApps.Metro can't use `Task.FromResult`. However, just returning `new System.Threading.Tasks.Task(() => { })` doesn't help, because it won't execute the task. Instead, you should use a `TaskCompletionSource`. Since both methods in question already ise a TCS, in the PR I moved their creations to an earlier point and re-used them.

I noticed this issue when I tried to open two dialog at once, a continuation for the returned tasks never run.


punker76:
thx @thoemmi that makes sense

PR #2006, opened 2015-06-26
FIX: issue with RangeSlider with binding and values between ticks

drayde:
Had an issue with a range slider where I could not move the thumb all the way to the right. In this case CalculateNextTick() moved into the !IsDoubleCloseToInt() code and that did not respect the passed in distance.

### DIFF
--- a/MahApps.Metro/Controls/RangeSlider.cs
+++ b/MahApps.Metro/Controls/RangeSlider.cs
@@ -2025,17 +2025,15 @@ namespace MahApps.Metro.Controls
             if (!IsMoveToPointEnabled)
             {
                 //Check if current value is exactly Tick value or it situated between Ticks
-                if (!IsDoubleCloseToInt((chekingValue - Minimum) / TickFrequency))
+                if (!IsDoubleCloseToInt((chekingValue + distance - Minimum) / TickFrequency))
                 {
-                    double x = (chekingValue - Minimum) / TickFrequency;
+                    double x = (chekingValue + distance - Minimum) / TickFrequency;
                     distance = TickFrequency * (int)x;
                     if (dir == Direction.Increase)
                     {
                         distance += TickFrequency;
                     }
                     distance = (distance - Math.Abs(chekingValue - Minimum));
-                    _currenValue = 0;
-                    return Math.Abs(distance);
                 }
             }
             //If we need move directly to next tick without calculating the difference between ticks
--- a/MahApps.Metro/Controls/RangeSlider.cs
+++ b/MahApps.Metro/Controls/RangeSlider.cs
@@ -2036,7 +2036,8 @@ namespace MahApps.Metro.Controls
                         distance += TickFrequency;
                     }
                     distance = (distance - Math.Abs(checkingValuePos));
-                }
+                    _currenValue = 0;
+                    return Math.Abs(distance);                }
             }
             //If we need move directly to next tick without calculating the difference between ticks
             //Use when MoveToPoint disabled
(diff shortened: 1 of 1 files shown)


MahApps.Metro/Controls/RangeSlider.cs
```diff
@@ -2025,17 +2025,15 @@ private Double CalculateNextTick(Direction dir, double chekingValue, double dist
             if (!IsMoveToPointEnabled)
             {
                 //Check if current value is exactly Tick value or it situated between Ticks
-                if (!IsDoubleCloseToInt((chekingValue - Minimum) / TickFrequency))
+                if (!IsDoubleCloseToInt((chekingValue + distance - Minimum) / TickFrequency))
```
shiftkey: While we're in here making this change, I have a couple of suggestions:

 - can we extract `(chekingValue + distance - Minimum)` into a local variable that matches up with that documented comment, for better readability - perhaps a name like `currentValueChanged`?
 - `chekingValue` should be renamed to `checkingValue` - that might have some flow-on effects...

MahApps.Metro/Controls/RangeSlider.cs
```diff
@@ -2025,17 +2025,15 @@ private Double CalculateNextTick(Direction dir, double chekingValue, double dist
...
                 {
-                    double x = (chekingValue - Minimum) / TickFrequency;
+                    double x = (chekingValue + distance - Minimum) / TickFrequency;
```
shiftkey: Following on from the earlier comment, we're re-calculating this value a second time. Could we simplify it further to reuse the last value?

punker76: :+1:

MahApps.Metro/Controls/RangeSlider.cs
```diff
@@ -2025,17 +2025,15 @@ private Double CalculateNextTick(Direction dir, double chekingValue, double dist
...
                     distance = TickFrequency * (int)x;
                     if (dir == Direction.Increase)
                     {
                         distance += TickFrequency;
                     }
                     distance = (distance - Math.Abs(chekingValue - Minimum));
-                    _currenValue = 0;
```
drayde: I was not sure about the return statement, I removed it and the algorithm now falls through to the other case. Seems to work as indended. What do you say? OK?

drayde:
Added your suggestions to the pull request. Do you need anything else so you can merge it?

MahApps.Metro/Controls/RangeSlider.cs
```diff
@@ -2020,22 +2020,22 @@ private void SnapToTickHandle(ButtonType type, Direction mDirection, Double diff
...
                 {
-                    double x = (chekingValue - Minimum) / TickFrequency;
                     distance = TickFrequency * (int)x;
                     if (dir == Direction.Increase)
                     {
                         distance += TickFrequency;
                     }
-                    distance = (distance - Math.Abs(chekingValue - Minimum));
-                    _currenValue = 0;
-                    return Math.Abs(distance);
```
punker76: why not returning here? is this right?

drayde: I am unsure. Isn't the code around line 2053 and onwards valid for this case as well?

punker76: @drayde i'm sure that we need this

without returning (the slider has a 20 tick frequency)
2015-07-01 22h28_03

with returning, which works correct
2015-07-01 22h30_01

drayde:
Should be ready to be merged now .-)

PR #2009, opened 2015-06-28
support for multiple dialogs (fixes #1974)

### LINKED ISSUE #1974: ShowMessageAsync above modal window
Hello,

I would like to use the ShowMessageAsync method to show a message when a modal dialog window is opened.
My possibilities:
1. Message window parent is the modal window. In this case the message is shown only in the smaller dialog window. Only the modal window goes darker.
2. Message window parent is the main application window. It is ok, the message is shown (going dark) all the full screen, but the dialog window is still white, and top, and hide the message window.
3. Use External dialog window -  Everything is ok, but in this case every message window is styled from external dialog window, but the "YesNo" message dialogs has another style (going darker).

I would like to use the same style for my application, so my best choice would be the 2. possibility, but the modal dialog hides my message. :(

Thanks in advance!

thoemmi:
This PR adds support for multiple dialogs at once as requested in #1974. There can only be one active dialog, and there's no support for switching between dialogs. It's like a stack of dialogs.

I hope the changes are self-explanatory.

However, there's one breaking change: previously the `MetroWindow`'s template contained a part named `PART_MetroDialogContainer`, now there are two parts named `PART_MetroActiveDialogContainer` and `PART_MetroInactiveDialogsContainer`. When someone has created his own style for `MetroWindow` it will break.

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/DialogManager.cs
+++ b/MahApps.Metro/Controls/Dialogs/DialogManager.cs
@@ -4,7 +4,6 @@ using System.Threading.Tasks;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Media;
-using MahApps.Metro.Behaviours;
 
 namespace MahApps.Metro.Controls.Dialogs
 {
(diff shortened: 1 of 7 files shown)


punker76:
@thoemmi i'm fine with this (little breaking) changes (maybe i must update my SimpleChildWindow). but you forgot the `CleanWindow`style ;-)

thoemmi:
Shame on me :flushed: I'll fix it.

punker76:
:+1:

IsNull:
Much appreciated enhancement :+1:

thoemmi:
Motivated by selfishness ^^

PR #2010, opened 2015-06-29
update pivot's selected-item-changed

ycaoyang:
update pivot's header selected item when modify the Selected-Pivot-Item by code, which was missing。


punker76:
fixed by 0c1df8ef91abae1201f52729dc5641221bce6f4e

PR #2019, opened 2015-07-05
Update MessageDialog.cs

bigworld12:


punker76:
@bigworld12 so https://github.com/MahApps/MahApps.Metro/pull/2022 will be a good solution for this

PR #2024, opened 2015-07-06
Implement ButtonsPlacement property to have possibility to set NUD on the left side

QuantumDeveloper:


MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -70,7 +70,13 @@ private static void IsReadOnlyPropertyChangedCallback(DependencyObject dependenc
             typeof(NumericUpDown),
             new FrameworkPropertyMetadata(default(double?), FrameworkPropertyMetadataOptions.BindsTwoWayByDefault, OnValueChanged, CoerceValue));
 
-        public static readonly DependencyProperty MinimumProperty = DependencyProperty.Register(
+      public static readonly DependencyProperty ButtonsPlacementProperty = DependencyProperty.Register(
+           "ButtonsPlacement",
+           typeof(Dock),
```
punker76: Can you create an enumeration for this, instead using `Dock`?

something like

```csharp
enum ButtonAlignment
{
    Left,
    Right
}
```

i think all other values make no sense

MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -70,7 +70,13 @@ private static void IsReadOnlyPropertyChangedCallback(DependencyObject dependenc
...
+           typeof(NumericUpDown),
+           new FrameworkPropertyMetadata(Dock.Right, FrameworkPropertyMetadataOptions.BindsTwoWayByDefault));
+
+      public static readonly DependencyProperty MinimumProperty = DependencyProperty.Register(
```
punker76: can you follow MahApps' coding styleguide?

punker76:
@QuantumDeveloper two notes...
1) why changing the buttons from "+ -" to "- +"
2) the ClearTextButton is still right...
image

QuantumDeveloper:
1. I made mirroring of the buttons to make it symmetric. If you want, I can change order.
2. Need to deal with it.

thoemmi:
Wouldn't be `ButtonsPlacement` a better name than `ButtonsPlacementVariant`? /cc @punker76

punker76:
@thoemmi what about `Location` for the enumeration and `ButtonLocation` for the dp? (about how xceed does it)

thoemmi:
Maybe `ButtonsAlignment` is more consistent with WPF's naming. And `...Buttons` should be plural, because it refers to all buttons of a TextBox.

punker76:
@thoemmi yeah, i forgot the `s` and i am agreed with `ButtonsAlignment`... for both? enum and dp?

thoemmi:
Yes, I'd call both the enum and the dp `ButtonsAlignment`.

punker76:
@thoemmi done 66e26652c249717ede763ae5cc842be3d78579f7

thoemmi:
:+1:

PR #2007, opened 2015-06-26
Allow titlebar to be realigned using new TitleAlignment property

edjubuh:
Added the ability to change the horizontal alignment of the title content using the TitleAlignment dependency property.

```XAML
<ma:MetroWindow
    ...
    TitleAlignment="Center" Title="Window Title"
>
   ...
</ma:MetroWindow>
```

image

PR #2040, opened 2015-07-17
Add TitleHorizontalAlignment property on the MetroWindow.

spasas:
Our designer requested the title to be shown on the left of the Window. It may be useful for other projects as well.


thoemmi:
This is similar to #2007. Change the `TitleTemplate` of your window.

spasas:
I agree that it is pretty close to #2007 almost the same even. It's your call, but it seems overkill to me to template the title if I only want it to be aligned left.

thoemmi:
**You** only want it to be left aligned. The next user might want to have bold text, another one italic. IMHO there are already way too many dependency properties on `MetroWindow`, and as shown in #2007 the title is really simple to style.

spasas:
Fair enough :) You can cancel the pull request then.

thoemmi:
@spasas don't get me wrong, your requirement is totally comprehensible. It's just that it would open a can of worms :wink:

Issue #1149, opened 2014-03-12
ProgressDialog cannot find the AccentColorBrush Resources


ye4241:
<ResourceDictionary.MergedDictionaries>
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Controls.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Fonts.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Colors.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Accents/Blue.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Accents/BaseLight.xaml" />
            </ResourceDictionary.MergedDictionaries>

Add all ResourceDictionary, still show cannot find the AccentColorBrush Resources when i used the MetroDialogColorScheme.Theme.
i try to track this error still no answer.Only follow shows in the debug when i try to close the progress dialog.
i've see the PART_ProgressBar, but i seems ok...please help me out,i need the progress dialog works.

>
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='MahApps.Metro.Controls.MetroWindow', AncestorLevel='1''. BindingExpression:Path=ActualWidth; DataItem=null; target element is 'MetroProgressBar' (Name='PART_ProgressBar'); target property is 'Width' (type 'Double')
System.Windows.Data Error: 4 : Cannot find source for binding with reference 'RelativeSource FindAncestor, AncestorType='MahApps.Metro.Controls.Dialogs.ProgressDialog', AncestorLevel='1''. BindingExpression:Path=ProgressBarForeground; DataItem=null; target element is 'MetroProgressBar' (Name='PART_ProgressBar'); target property is 'Foreground' (type 'Brush')
System.Windows.Media.Animation Warning: 6 : Unable to perform action because the specified Storyboard was never applied to this object for interactive control.; Action='Remove'; Storyboard='System.Windows.Media.Animation.Storyboard'; Storyboard.HashCode='41580444'; Storyboard.Type='System.Windows.Media.Animation.Storyboard'; TargetElement='System.Windows.Media.Animation.Storyboard'; TargetElement.HashCode='41580444'; TargetElement.Type='System.Windows.Media.Animation.Storyboard'
[...]

PR #2087, opened 2015-08-22
Fix obj ref null exception if there is no Window Theme at runtime.

ButchersBoy:
There isn't a Window Theme if using Material Design In Xaml Toolkit as the App.xaml is configured differently.  Safely jumping out at this point allows dialogs to work with Material Design In Xaml Toolkit.

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/BaseMetroDialog.cs
@@ -118,7 +118,7 @@ namespace MahApps.Metro.Controls.Dialogs
             {
                 var windowTheme = DetectTheme(this);
 
-                if (System.ComponentModel.DesignerProperties.GetIsInDesignMode(this) && windowTheme == null)
+                if (windowTheme == null)
                 {
                     return;
                 }


ButchersBoy:
part 2...if we allow suppression of the default resources I can allow Material Design styles to cascade through...

punker76:
I have another solution for this, cause not adding the `BaseMetroDialog` resource dictionary will get possible bugs (and warnings) with the dialogs itself (#2032, #1149).

The `BaseMetroDialog` holds some resource keys which are used in this dialogs styles.

So here is my proposal, just create a list of custom resource dictionaries in den `DialogSettings`, so the user can add additional resources (like the Material Design resources).

```csharp
public class MetroDialogSettings
{
        /// <summary>
    /// Gets/sets a custom resource dictionaries which can contains custom styles, brushes or something else.
    /// </summary>
    public List<ResourceDictionary> CustomResources { get; set; }
}
```

```csharp
public abstract class BaseMetroDialog : ContentControl
{
    private void Initialize()
    {
    	if (DialogSettings != null && !DialogSettings.SuppressDefaultResources)
    	{
    		this.Resources.MergedDictionaries.Add(new ResourceDictionary { Source = new Uri("pack://application:,,,/MahApps.Metro;component/Styles/Controls.xaml") });
    	}

        // add default resources, cause without you could get resource key not found warnings
        this.Resources.MergedDictionaries.Add(new ResourceDictionary { Source = new Uri("pack://application:,,,/MahApps.Metro;component/Styles/Fonts.xaml") });
        this.Resources.MergedDictionaries.Add(new ResourceDictionary { Source = new Uri("pack://application:,,,/MahApps.Metro;component/Styles/Colors.xaml") });
        this.Resources.MergedDictionaries.Add(new ResourceDictionary { Source = new Uri("pack://application:,,,/MahApps.Metro;component/Themes/Dialogs/BaseMetroDialog.xaml") });

        // just add some custom resources
        if (DialogSettings != null && DialogSettings.CustomResources != null)
        {
            foreach(var rd in DialogSettings.CustomResources)
            {
            	this.Resources.MergedDictionaries.Add(rd);
            }
        }

```

```csharp
        private void HandleTheme()
        {
            if (DialogSettings != null)
            {
                var windowTheme = DetectTheme(this);

                // no theming in design mode or no window theme was found
                // maybe we must give the users a option to handle the MetroDialogColorScheme by itself
                if (System.ComponentModel.DesignerProperties.GetIsInDesignMode(this) || windowTheme == null)
                {
                    return;
                }
```
@ButchersBoy What do you think?

ButchersBoy:
OK, I understand the error of my ways, so yes, that works for more, although I don'y need the extra resources list, I can make do with the single one you already have...I've simplified my PR.  Now we  can have even more integration :smile:  :

image

punker76:
@ButchersBoy GREAT! :+1: :trophy:

PR #2097, opened 2015-08-27
Progressdialog improvements

thoemmi:
As far as I can see there is some, er, room for improvements in `ProgressDialogController`:

* `ProgressDialogController.IsOpen` is only set in the c'tor and when the dialog is closed. In fact, I see no reason to have that property, so I eliminated it.
* By default the state of the progress bar in the dialog is indeterminated. IMHO this should be set by the user explicitly.
* `ProgressDialogController` makes direct calls to several controls of `ProgressDialog`, e.g. `WrappedDialog.PART_ProgressBar.IsIndeterminate = true`. Instead I'd like to add appropriate methods to `ProgressDialog`, because the controls are implementation details of the dialog and not the concern of the controller.

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/ProgressDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/ProgressDialog.cs
@@ -29,6 +29,7 @@ namespace MahApps.Metro.Controls.Dialogs
                 ProgressBarForeground = Brushes.White;
             }
         }
+
         internal ProgressDialog(MetroWindow parentWindow)
             : this(parentWindow, null)
         { }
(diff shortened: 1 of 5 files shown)


bigworld12:
I totally disagree on the first commit
i use multi threading in my program which means that i can accidentally
open 2 dialogs at the same time
So i need to check the IsOpen property
What i would suggest:
1- add a closed event to the controller but leave the isopen property and
pass a parameter called result that will be an enum (check suggest 3)

2- add a property to make the dialogue access the ui controls so you can
change style, template,etc...

3-add a room for two auxiliary buttons like in normal dialogues with the
ability to hide/show them in the controller

thoemmi:
@bigworld12 Thanks for your feedback.

> i use multi threading in my program which means that i can accidentally
open 2 dialogs at the same time
So i need to check the IsOpen property

Ok, I should have inspected the API more precisely. Yes, an `Closed` would be helpful.

About changing style, additional buttons, etc: I'd leave ProgressDialog as is. If someone wants to customize the dialog, he still can derive his own implementation from `BaseMetroDialog`.

punker76:
@thoemmi

> About changing style, additional buttons, etc: I'd leave ProgressDialog as is. If someone wants to customize the dialog, he still can derive his own implementation from BaseMetroDialog.

:+1:

thoemmi:
@punker76 When I add a `Closed` event to `ProgressDialogController` is this PR acceptable? (the removal of `IsOpen` is a breaking change though)

thoemmi:
I recreated the `IsOpen` property and added a `Closed` event to `ProgressDialogController`. IMHO this PR is ready to be merged :grin:

PR #2112, opened 2015-09-04
If the user.config file is corrupt the whole application crashes

Remolutionary:
I had the problem, that a customer could suddenly not open my
application anymore. It turned out that this is because the user.config
file was just empty and this caused a ConfigurationErrorsException. Then I
added an UnhandledExceptionHandler but this exception seems to happen
earlier. The only working solution is to delete the odd user.config and
restart the application.
For more details see here:
http://www.codeproject.com/Articles/30216/Handling-Corrupt-user-config-Settings

You can test this by editing the user.config and delete all content. But do NOT delete the file!
Then you need a window with the setting SaveWindowPosition="True" - otherwise no user.config file is created.

### DIFF
--- a/MahApps.Metro/Controls/WindowSettings.cs
+++ b/MahApps.Metro/Controls/WindowSettings.cs
@@ -2,7 +2,6 @@
 using System.ComponentModel;
 using System.Configuration;
 using System.Diagnostics;
-using System.IO;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Interop;
--- a/MahApps.Metro/Controls/WindowSettings.cs
+++ b/MahApps.Metro/Controls/WindowSettings.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.Configuration;
 using System.Diagnostics;
+using System.IO;
 using System.Runtime.InteropServices;
 using System.Windows;
 using System.Windows.Interop;
@@ -64,9 +65,20 @@ namespace MahApps.Metro.Controls
         {
             get
             {
-                if (this["UpgradeSettings"] != null)
+                try
                 {
-                    return (bool)this["UpgradeSettings"];
+                    if (this["UpgradeSettings"] != null)
+                    {
+                        return (bool)this["UpgradeSettings"];
+                    }
+                }
+                catch (Exception ex)
+                {
+                    Debug.WriteLine("Failed to load settings file:\r\n{0}", ex);
+                    var filename = ((ConfigurationErrorsException)ex.InnerException).Filename;
+                    File.Delete(filename);
+                    Process.Start(Application.ResourceAssembly.Location);
+                    Application.Current.Shutdown();
                 }
                 return true;
             }
--- a/MahApps.Metro/Controls/WindowSettings.cs
+++ b/MahApps.Metro/Controls/WindowSettings.cs
@@ -72,13 +72,10 @@ namespace MahApps.Metro.Controls
                         return (bool)this["UpgradeSettings"];
                     }
                 }
-                catch (Exception ex)
+                catch (ConfigurationErrorsException ex)
                 {
-                    Debug.WriteLine("Failed to load settings file:\r\n{0}", ex);
                     var filename = ((ConfigurationErrorsException)ex.InnerException).Filename;
-                    File.Delete(filename);
-                    Process.Start(Application.ResourceAssembly.Location);
-                    Application.Current.Shutdown();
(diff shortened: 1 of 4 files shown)


Remolutionary:
An exception is already thrown. The isssue has been that I could not catch it in my application but it turned out that I defined my UnhandledExepctionHandler too late. I'm using NLOG as log System and the ConfigurationErrorException is thrown when  a logger is defined:
 public App()
        {
                _logger = LogManager.GetLogger("App"); 
        }
If I add a trycatch to this line I can catch the error although the logger does not cause the problem. In fact it is the settings-system which loads the settings as soon as the app starts.

So I think my pull request is in fact needless...sorry.

Remolutionary:
One thing I forgot: 
The ConfigurationErrorsException can't be catched by an UnhandledExceptionHandler. So it is necessary to wrap it into a MahAppsException. Otherwise the user can't handle the situation because he does not know when this error arises. 

In my special case with the logger I can catch the exception now but this is more accidentally I would say. If I remove this line of code I have no chance to catch this kind of exception.

So I reopen my pull request.

MahApps.Metro/Controls/WindowSettings.cs
```diff
@@ -64,9 +64,17 @@ public bool UpgradeSettings
...
+                    if (this["UpgradeSettings"] != null)
+                    {
+                        return (bool)this["UpgradeSettings"];
+                    }
+                }
+                catch (ConfigurationErrorsException ex)
+                {
+                    var filename = ((ConfigurationErrorsException)ex.InnerException).Filename;
+                    throw new MahAppsException(string.Format("The settings file {0} seems to be corrupted", filename), ex);
                 }
```
punker76: sorry for my nitpicking ;-) but why must the inner exception cast to ConfigurationErrorsException? the ex is already a ConfigurationErrorsException?

should this better so?

```csharp
catch (ConfigurationErrorsException ex)
{
    var filename = ex.Filename;
    throw new MahAppsException(string.Format("The settings file {0} seems to be corrupted", filename), ex.InnerException);
}
```

Remolutionary: Yes they are the same exception type but in the outer exception Filename is null. You really have to go to the InnerException and cast it… I just debugged the code again to be sure about that.

 

Von: Jan Karger [mailto:notifications@github.com] 
Gesendet: Freitag, 4. September 2015 21:24
An: MahApps/MahApps.Metro <MahApps.Metro@noreply.github.com>
Cc: Markus <markus.remmel@remolution-software.com>
Betreff: Re: [MahApps.Metro] If the user.config file is corrupt the whole application crashes (#2112)

 

In MahApps.Metro/Controls/WindowSettings.cs <https://github.com/MahApps/MahApps.Metro/pull/2112#discussion_r38785880> :

>                  {
> -                    return (bool)this["UpgradeSettings"];
> +                    if (this["UpgradeSettings"] != null)
> +                    {
> +                        return (bool)this["UpgradeSettings"];
> +                    }
> +                }
> +                catch (ConfigurationErrorsException ex)
> +                {
> +                    var filename = ((ConfigurationErrorsException)ex.InnerException).Filename;
> +                    throw new MahAppsException(string.Format("The settings file {0} seems to be corrupted", filename), ex);
>                  }

sorry for my nitpicking ;-) but why must the inner exception cast to ConfigurationErrorsException? the ex is already a ConfigurationErrorsException?

should this better so?

catch (ConfigurationErrorsException ex)
{
    var filename = ex.Filename;
    throw new MahAppsException(string.Format("The settings file {0} seems to be corrupted", filename), ex.InnerException);
}

punker76: @Remolutionary ok, really don't know this, so i think this could be a final solution

```cshap
catch (ConfigurationErrorsException ex)
{
    var innerEx = ex.InnerException as ConfigurationErrorsException;
    var filename = innerEx != null ? innerError.Filename : "<unknown>";
    throw new MahAppsException(string.Format("The settings file '{0}' seems to be corrupted", filename), ex.InnerException);
}
```

thx

thoemmi: I'd prefer more robust code like this:
```
catch (ConfigurationErrorsException ex)
{
    string filename = null;
    while (ex != null && (filename = ex.Filename) == null)
    {
        ex = ex.InnerException as ConfigurationErrorsException;
    }
    throw new MahAppsException(string.Format("The settings file '{0}' seems to be corrupted", filename ?? "<unknown>"), ex);
}
```

punker76: @thoemmi :+1:

PR #2143, opened 2015-09-23
[WIP][RFC] Fix for Glow corners

### LINKED ISSUE #2118: Glow border is missing the corners
It's a little bit subtle, but you can see in this screenshot that there's no glow border at the corner (and the same is true for the other 3 corners). Is that as-designed, or a bug? It looks pretty odd, and isn't what the "standard" Windows glow looks like.

mahapps-glow

punker76:
So, this is my attempt to fix the glow corners.

Current situation (before this PR):

image

image

And now (with this PR):

image

image

image

- [x] rewrite glow templates (don't use margin, cause inner border flickers in horizontal and vertical direction on resizing, margins... ts)
- [x] fix show non active glow border brush

Closes #2118 Glow border is missing the corners

/cc @thoemmi @flagbug


thoemmi:
Looks good on Windows 10:
image

It's hard to notice on a black background, but I'm fine with that:
image
Inactive state is more recognizable:
image

PR #2144, opened 2015-09-28
Fix incorrect restore location after Aero Snap

tgjones:
Fixes an odd side effect in calling GetWindowPlacement from WindowChromeWorker._HandleGetMinMaxInfo that caused the window to restore to the end of the drag instead of the beginning after doing an Aero Snap.

Steps to repro the bug that this PR fixes:

1. In the MahApps demo app, drag window to top of screen, which maximises the window.
2. Click Restore.
3. Window is restored to a location slightly off the top of the screen (which is, incorrectly, the location of the end of the previous drag).

Calling `IsZoomed` instead of `GetWindowPlacement` fixes this issue.


punker76:
@tgjones thx, this works really better.

tgjones:
Thanks for merging :)

(I'm really curious what the history is behind the naming of that API - why is it `IsZoomed`, not `IsMaximized`...)

PR #2150, opened 2015-10-02
Some progressbar dialog fixes

bigworld12:
i think this is how progressbar dialog should be , removed the obsolete from IsOpen property and made it read only , also added cancelled event

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
+++ b/MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
@@ -22,7 +22,7 @@ namespace MahApps.Metro.Controls.Dialogs
         /// <summary>
         /// This event is raised when the associated <see cref="ProgressDialog"/> was cancelled by the user.
         /// </summary>
-        public event EventHandler Cancelled;
+        public event EventHandler Canceled;
 
         /// <summary>
         /// Gets if the Cancel button has been pressed.
@@ -55,7 +55,7 @@ namespace MahApps.Metro.Controls.Dialogs
         {
             Action action = () => {
                 IsCanceled = true;
-                var handler = Cancelled;
+                var handler = Canceled;
                 if (handler != null)
                 {
                     handler(this, EventArgs.Empty);
--- a/MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
+++ b/MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
@@ -27,21 +27,19 @@ namespace MahApps.Metro.Controls.Dialogs
         /// <summary>
         /// Gets if the Cancel button has been pressed.
         /// </summary>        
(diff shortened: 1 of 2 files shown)


MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
```diff
@@ -12,23 +12,36 @@ public class ProgressDialogController
...
+        /// Gets if the Cancel button has been pressed.
+        /// </summary>        
+        public bool IsCancelled {get{ return m_iscanceled;}}
+        private bool m_iscanceled;
+
+        /// <summary>
+        /// Gets if the wrapped ProgressDialog is open.
+        /// </summary>        
+        public bool IsOpen { get { return m_isopen; } }
+        private bool m_isopen;
```
punker76: why this change? IsOpen is already marked as obsolete, so this is not necessary...

bigworld12: i use IsOpen to check if the dialog is visible so i close it programmatically , so it filled vs with warnings of it being obsolete

MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
```diff
@@ -12,23 +12,36 @@ public class ProgressDialogController
...
+   
+        /// <summary>
+        /// This event is raised when the associated <see cref="ProgressDialog"/> was cancelled by the user.
+        /// </summary>
+        public event EventHandler Cancelled;
+
+        /// <summary>
+        /// Gets if the Cancel button has been pressed.
+        /// </summary>        
+        public bool IsCancelled {get{ return m_iscanceled;}}
```
punker76: this can be simplyfied to `public bool IsCancelled { get; private set; }`
and i would prefere `IsCanceled`

bigworld12: ok i will bring it back

MahApps.Metro/Controls/Dialogs/ProgressDialogController.cs
```diff
@@ -12,16 +12,27 @@ public class ProgressDialogController
...
         /// <summary>
-        /// This event is raised when the associated <see cref="ProgressDialog"/> was closed.
+        /// This event is raised when the associated <see cref="ProgressDialog"/> was closed programmatically.
         /// </summary>
         public event EventHandler Closed;
+   
+        /// <summary>
+        /// This event is raised when the associated <see cref="ProgressDialog"/> was cancelled by the user.
+        /// </summary>
+        public event EventHandler Cancelled;
```
punker76: ;-) -> `Canceled`

remcoros: Typical Americanism. Everywhere else in the world it's cancelled (with double l).

http://grammarist.com/spelling/cancel/

PR #2171, opened 2015-10-20
Change Slider value on MouseWheel event.

xxMUROxx:
Possibility to rotate the mouse wheel and change the value of the slider in following situations:
    * None
    * Only when the slider is focuesed and the mouse wheel is rotated
    * If the mouse pointer is over the slider and the mouse wheel is rotated

### DIFF
--- a/MahApps.Metro/Controls/Helper/SliderHelper.cs
+++ b/MahApps.Metro/Controls/Helper/SliderHelper.cs
@@ -1,0 +1,84 @@
+﻿namespace MahApps.Metro.Controls
+{
+    using System.Windows;
+    using System.Windows.Controls;
+    using System.Windows.Input;
+
+    public class SliderHelper
+    {
+        public static readonly DependencyProperty ChangeValueByProperty = DependencyProperty.RegisterAttached(
+            "ChangeValueBy",
+            typeof(MouseWheelChange),
+            typeof(SliderHelper),
+            new PropertyMetadata(MouseWheelChange.SmallChange));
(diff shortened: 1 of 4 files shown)


MahApps.Metro/Controls/Helper/SliderHelper.cs
```diff
@@ -0,0 +1,84 @@
...
+            if ((MouseWheelState)e.NewValue != MouseWheelState.None)
+            {
+                slider.Unloaded += OnUnloaded;
+                slider.PreviewMouseWheel += OnPreviewMouseWheel;
+            }
+            else
+            {
+                UnregisterEvents(slider);
+            }
+        }
```
punker76: we should always unregister the events

```csharp
if (e.NewValue != e.OldValue)
{
	var slider = (Slider)d;
	UnregisterEvents(slider);
	if ((MouseWheelState)e.NewValue != MouseWheelState.None)
	{
	    slider.Unloaded += OnUnloaded;
	    slider.PreviewMouseWheel += OnPreviewMouseWheel;
	}
}
```

thoemmi:
BTW, I would introduce `RegisterEvents(Slider slider)`. Though it would be called only from one location, you would get symmetric methods, `RegisterEvents` and `UnregisterEvents`.

punker76:
@thoemmi :+1:

xxMUROxx:
@thoemmi @punker76 Done :+1:

khteh:
Is there any change completion event for this mousewheel change of the slider value?

PR #2172, opened 2015-10-20
Custom input dialog

jords1987:
Work in progress.. wondering, am I going the right way about this? Also, awaiting using ".Result" is not working, All feedback welcomed


jords1987:
Appveyor keeps gailin

jords1987:
*failing

jords1987:
Taken out 'using' statements that are not required

PR #2176, opened 2015-10-23
Custom input dialog

jords1987:
Allows for creation of an input dialog that returns a custom object (MVVM supported)

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/CustomInputDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/CustomInputDialog.cs
@@ -13,7 +13,6 @@ namespace MahApps.Metro.Controls.Dialogs
         }
 
         public static readonly DependencyProperty MessageProperty = DependencyProperty.Register("Message", typeof(string), typeof(CustomInputDialog<T>), new PropertyMetadata(default(string)));
-        public static readonly DependencyProperty InputProperty = DependencyProperty.Register("Input", typeof(TaskCompletionSource<T>), typeof(CustomInputDialog<T>), new PropertyMetadata(default(T)));
         public static readonly DependencyProperty AffirmativeButtonTextProperty = DependencyProperty.Register("AffirmativeButtonText", typeof(string), typeof(CustomInputDialog<T>), new PropertyMetadata("OK"));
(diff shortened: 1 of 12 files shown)


samples/MetroDemo/MainWindowViewModel.cs
```diff
@@ -339,7 +340,32 @@ public ICommand ShowProgressDialogCommand
...
+        private async void RunCustomInputFromVm()
+        {
+            var customDialog = new PersonInputDialog() {Title = "Select Person"};
+
+            //var resultingPerson = _dialogCoordinator.ShowCustomInputAsync(this, customDialog).Result; //todo: this doesnt work
+            var resultingPerson = await _dialogCoordinator.ShowCustomInputAsync(this, customDialog);
+
+            if (resultingPerson != null) {
+                await _dialogCoordinator.ShowMessageAsync(this, "YOUR CHOICE",
+                    $"You selected {resultingPerson.Name}, who is aged {resultingPerson.Age}");
```
punker76: this is a c# 6.0 feature which is currently not supported here in MahApps solution

jords1987:
Removed the c#6 features now, thanks

MahApps.Metro/Controls/Dialogs/CustomInputDialog.cs
```diff
@@ -0,0 +1,39 @@
...
+        public abstract Task<T> WaitForButtonPressAsync(); 
+
+        protected override void OnLoaded()
+        {
+            this.AffirmativeButtonText = this.DialogSettings.AffirmativeButtonText;
+            this.NegativeButtonText = this.DialogSettings.NegativeButtonText;
+        }
+
+        public static readonly DependencyProperty MessageProperty = DependencyProperty.Register("Message", typeof(string), typeof(CustomInputDialog<T>), new PropertyMetadata(default(string)));
+        public static readonly DependencyProperty InputProperty = DependencyProperty.Register("Input", typeof(TaskCompletionSource<T>), typeof(CustomInputDialog<T>), new PropertyMetadata(default(T)));
```
punker76: @jords1987 there is no property for `InputProperty`? and is `typeof(TaskCompletionSource<T>)` the correct value for the type?

jords1987:
Thanks... this was from how I first tried to implement the functionality... I missed the removal. I also missed removing the textbox from the input dialog I copied so corrected that... let me know if there is anything else

Issue #2149, opened 2015-09-30
NumericUpDown Culture problem


fredericoregateiro:
Hello,

I'm using the NumericUpDown control in a app with the culture pt-PT, in this culture the decimal point is ',' and using a numeric key pad the decimal key is ignored.

When the user press the decimal key, even when the decimal point is not the dot, the text entered should be the given culture decimal separator, like System.Globalization.CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator.

Is any way to do this, or is this a limitation of the current control.

I've tested with the 1.1.2 and 1.1.3-ALPHA174 versions, in the MetroDemo.NET45 project.

Thanks,
Frederico

punker76:
@fredericoregateiro it seems you must set the Culture explicit https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Controls/NumericUpDown.cs#L294
cause NumericUpDown uses `Language.GetSpecificCulture();` by default https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Controls/NumericUpDown.cs#L442
NumericUpDown check and use `NumberDecimalSeparator` already 
https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Controls/NumericUpDown.cs#L651

fredericoregateiro:
@punker76 i have the culture set on the control also
Culture="pt-PT"
With this culture, the decimal point on my keypad doesn't work, only the comma on the qwerty keyboard sets the decimal point correctly. The decimal key works only with the culture set like "en-US" or similar.

Using the pt-PT culture, the users expect that the decimal point on the keypads set the comma ',' and start to fill the decimal places.

Thanks,
Frederico

punker76:
@fredericoregateiro how is your os keyboard configured?

fredericoregateiro:
Is this picture enough?

![settings](https://cloud.githubusercontent.com/assets/6304701/10540914/75173296-7404-11e5-9776-cadcf67da561.jpg)

The keypad i'm using is the numeric keypad, the most used to enter numeric values. Like this one:

![asus_n550_15](https://cloud.githubusercontent.com/assets/6304701/10540993/52188938-7405-11e5-950f-5b65a19523c3.jpg)

punker76:
question is, what comes in `OnPreviewTextInput` if you hit the numeric delete key?

```csharp
[...]

PR #2199, opened 2015-11-05
 NumericUpDown Culture problem #2149

fredericoregateiro:
Tested with different cultures and it seems to work just fine.
When the culture NumberDecimalSeparator is ',' (comma) the control replaces the '.' (period or decimal key) with the corrent NumberDecimalSeparator symbol.

All of the remaining functionality is the same.

### DIFF
--- a/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/MahApps.Metro/Controls/NumericUpDown.cs
@@ -1037,20 +1037,17 @@ namespace MahApps.Metro.Controls
             {
                 TextBox textBox = sender as TextBox;
 
-                if (this.Culture != null)
+                if (textBox.Text.Contains(this.SpecificCultureInfo.NumberFormat.NumberDecimalSeparator) == false)
                 {
-                    if (textBox.Text.Contains(this.Culture.NumberFormat.NumberDecimalSeparator) == false)
-                    {
-                        //the control doesn't contai the decimal separator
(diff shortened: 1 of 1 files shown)


fredericoregateiro:
@punker76 do you think this can get on the 1.2.0 release?

MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -1032,6 +1032,26 @@ private void EnableDisableUpDown()
...
             _manualChange = true;
+
+            if (e.Key == Key.Decimal || e.Key == Key.OemPeriod)
+            {
+                TextBox textBox = sender as TextBox;
+
+                if (this.Culture != null)
+                {
+                    if (textBox.Text.Contains(this.Culture.NumberFormat.NumberDecimalSeparator) == false)
+                    {
```
punker76: It's better to use `SpecificCultureInfo` property nstead `Culture`.

punker76:
@fredericoregateiro I will not decide this alone...

/cc @xxMUROxx

fredericoregateiro:
So, any decision on this one?

xxMUROxx:
@fredericoregateiro sorry, I have forgotten to test this. I will give you then a feedback

xxMUROxx:
@punker76 I have tested it. I have found no problems.
In my opinion we can merge this PR

PR #2222, opened 2015-11-20
Pad out the Custome-Dialog-via-MVVM example

ButchersBoy:
Saw a few people have been struggling with this in the forum so padded out the example to show how it can be done.  There's no changes to the core .dll, pure demo stuff.


punker76:
@ButchersBoy thx, mate :+1:

PR #2226, opened 2015-11-23
Use thumb for dragmove

### LINKED ISSUE #1635: Fylout window not draggable?
When I open a flyout over the whole window, the window itself is not draggable using the title bar of the window. Is this a bug or can I enable this?

### LINKED ISSUE #1849: System menu present everywhere when fullscreen is toggled on
When having a MetroWindow in fullscreen you end up with the titlebar system menu on right click (Restore, minimize...) that appears everywhere on the application.

To test this behavior you can either use the MahApps metro demo application or just create a new MetroWindow and set the following properties:
IgnoreTaskbarOnMaximize="True"
UseNoneWindowStyle="True"
WindowState="Maximized"
Now if you right click anywhere in the window you will see the title bar system menu.

You can avoid this problem by setting the ShowSystemMenuOnRightClick property to False but I am guessing it is not a correct default behavior for the window, what do you think?

### LINKED ISSUE #2036: Avalondock Anchorables could not be dragged inside MetroWindow
HI. I noticed that when Iam  using MetroWindow with avalondock, I cannot drag AnchorableGroup out of the WIndow. It just moves th whole window itself. But with standart WPF window, everything working correct.

Could this issue be fixed somehow?

For me its critical, because it affect features of docking control.

### LINKED ISSUE #2122: Use behavior/attached property to control window dragging
I have a `MetroWindow` with custom title bar by setting `ShowTitleBar="False"` and `IsWindowDraggle="False"`, but I also want this custom title bar to be draggable. It would be nice if there is a helper behavior that once attached, mouse down on the associated object would start dragging the window as this approach is more flexible and reusable. What do you think?

punker76:
- [x] use thumb for DragMove actions (makes it possible that animations not hangs)
- [x] double click in thumb
- [x] DragMove with UseNoneWindowStyle
- [x] right click on thumb
- [x] allow DragMove with open Flyouts
- [x] some code cleanup after Flyout implementation
- [x] test it with Avalon dock for @QuantumDeveloper 
- [x] hopefully not a breaking change :-D

thx @ButchersBoy for the code from [Dragablz](https://github.com/ButchersBoy/Dragablz)

Closes #1821 Move window when flyout is open
Closes #1635 Fylout window not draggable?
Closes #2036 Avalondock Anchorables could not be dragged inside MetroWindow
Closes #2122 Use behavior/attached property to control window dragging
Closes #1849 System menu present everywhere when fullscreen is toggled on

### DIFF
--- a/MahApps.Metro/Controls/Flyout.cs
+++ b/MahApps.Metro/Controls/Flyout.cs
@@ -1,7 +1,7 @@
 ﻿using System;
-using System.Linq;
 using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Controls.Primitives;
 using System.Windows.Input;
 using System.Windows.Media;
 using System.Windows.Media.Animation;
@@ -15,6 +15,7 @@ namespace MahApps.Metro.Controls
     /// <seealso cref="FlyoutsControl"/>
     /// </summary>
     [TemplatePart(Name = "PART_BackButton", Type = typeof(Button))]
+    [TemplatePart(Name = "PART_WindowRestoreThumb", Type = typeof(Thumb))]
     [TemplatePart(Name = "PART_Header", Type = typeof(ContentPresenter))]
     [TemplatePart(Name = "PART_Content", Type = typeof(ContentPresenter))]
     public class Flyout : ContentControl
@@ -212,6 +213,13 @@ namespace MahApps.Metro.Controls
             this.Loaded += (sender, args) => UpdateFlyoutTheme();
         }
 
+        private MetroWindow parentWindow;
+
+        private MetroWindow ParentWindow
+        {
+            get { return this.parentWindow ?? (this.parentWindow = this.TryFindParent<MetroWindow>()); }
+        }
+
         private void UpdateFlyoutTheme()
         {
             var flyoutsControl = this.TryFindParent<FlyoutsControl>();
@@ -221,7 +229,7 @@ namespace MahApps.Metro.Controls
                 this.Visibility = flyoutsControl != null ? Visibility.Collapsed : Visibility.Visible;
             }
 
-            var window = this.TryFindParent<MetroWindow>();
+            var window = this.ParentWindow;
             if (window != null)
             {
                 var windowTheme = DetectTheme(this);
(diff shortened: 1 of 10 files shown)


punker76:
@thoemmi If you have some time... the Flyout stuff is implemented and testable...

/cc @hercmaroudas @QuantumDeveloper

thoemmi:
@punker76 Here are my annotations:

* I think the name `windowRestoreThumb` (and `PART_WindowRestoreThumb`) quite confusing. This isn't about restoring the window but moving it, is it?

* Is the first condition in `MetroWindow.DoWindowMoveThumbOnDragDelta` correct? If `window.IsWindowDraggable` is `false` it should always return despite the properties of `dragDeltaEventArgs`.

* The `InvisibleThumbStyle` style is declared twice, in `MetroWindow.xaml` and `Flyout.xaml`. Mabe you should move to a central place, so there's only one declaration.

punker76:
@thoemmi 

> I think the name `windowRestoreThumb` (and `PART_WindowRestoreThumb`) quite confusing. This isn't about restoring the window but moving it, is it?

I was also rethinking about the name, maybe `PART_WindowTitleThumb` is better for this...

> Is the first condition in `MetroWindow.DoWindowMoveThumbOnDragDelta` correct? If `window.IsWindowDraggable` is `false` it should always return despite the properties of `dragDeltaEventArgs`.

Yeah should be!

> The `InvisibleThumbStyle` style is declared twice, in `MetroWindow.xaml` and `Flyout.xaml`. Mabe you should move to a central place, so there's only one declaration.

I'll do this.

thy @thoemmi

punker76:
@thoemmi all annotations done

thoemmi:
:+1: :shipit:

punker76:
:sunglasses:

ButchersBoy:
Found a bug with this on a 3 monitor setup; all of different resolutions and orientations...small PR incoming...

alexeytcherniak:
FYI, if ShowTitleBar="False", window dragging will not work, because window content  (MetroContentControl) overlays Title bar Thumb (PART_WindowTitleThumb). The problem is that the Grid row, where Thumb is placed, has Height=0 in this case, even if there are Minimize/Maximize/Close buttons visible. Thus, if ShowTitleBar="False", there is no way to move the window by dragging.

PR #2230, opened 2015-11-26
Hide username on LoginDialog

### LINKED ISSUE #2229: Hide username from LoginDialog
I want to be able to ask for just a password. The `LoginDialogSettings` could be extended with a property `UsernameVisibility` that controls the visibility of the Username `TextBox`. Default value would be `Visible`.

mgnslndh:
The LoginDialog can now be shown like this to hide the username and only ask for a password:

```
var result = await this.ShowLoginAsync("Authentication", "Enter your password", 
                                       new LoginDialogSettings 
                                       { 
                                           UsernameVisibility = Visibility.Collapsed
                                       });
```

Closes #2229

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/LoginDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/LoginDialog.cs
@@ -10,7 +10,7 @@ namespace MahApps.Metro.Controls.Dialogs
         private const string DefaultUsernameWatermark = "Username...";
         private const string DefaultPasswordWatermark = "Password...";
         private const Visibility DefaultNegativeButtonVisibility = Visibility.Collapsed;
-        private const Visibility DefaultUsernameVisibility = Visibility.Visible;
+        private const bool DefaultShouldHideUsername = false;
         private const bool DefaultEnablePasswordPreview = false;
 
         public LoginDialogSettings()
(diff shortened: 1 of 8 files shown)


punker76:
@mgnslndh thx, what is if you set to`Visibility=Hidden`? then you got a empty line. is this good? maybe a short bool like `HideUsername` could be better. what do you think?
/cc @thoemmi

mgnslndh:
Good point. I went for Visibility mostly because NegativeButton used it. But that does not mean it is appropriate for the Username of course. 

HideUsername is better. True value results in Collapsed visibility, otherwise Visible.

mgnslndh:
@punker76 HideUsername would require a custom BooleanToVisibilityConverter that inverts conversion (true => Collapse and false => Visible).

Is this OK, or should we use another property name like ShowUsername or IsUsernameVisible instead? 

One solution could be to use HideUsername in the LoginDialogSetting but use Visibility for the dependency property. What do you think?

thoemmi:
`LoginDialogSettings` should not have a property of type `Visibility`, that's an implementation detail. Use `bool` instead.

In general flags should be `false` by default, so `IsUsernameHidden'  or `ShouldHideUsername` would be approriate. If that requires a new `IConverter`, well, it's easily written.

`Collapsed` is better than `Hidden` so no screen estate is preserved.

punker76:
@mgnslndh can you change your PR to the suggestions?

mgnslndh:
@punker76 @thoemmi I can change the PR and use a `bool` for the property on `LoginDialogSettings`, but I'd like your feedback on the dependency property type. Should it also be `bool` like on the `LoginDialogSettings` property or can it be of the type `Visibility` in which case we do not need a `IValueConverter`.

mgnslndh:
@punker76 I've addressed the suggestions. Does it look OK?

punker76:
@mgnslndh i merged your changes, but removed your converter and uses the new `VisibilityHelper` property `IsCollapsed`.

/cc @thoemmi

Issue #2235, opened 2015-11-27
UseFloatingWatermark is not working on ComboBox


amghw:
```xaml
<ComboBox Controls:TextBoxHelper.Watermark="Country" Controls:TextBoxHelper.UseFloatingWatermark="True" DisplayMemberPath="country" 
IsEditable="True" ItemsSource="{Binding Countries}" />
```

Thanks for the wonderful toolkit!

punker76:
fixed with https://github.com/MahApps/MahApps.Metro/pull/2236

amghw:
Thanks!

PR #2236, opened 2015-11-27
Floating watermarks now trigger better

### LINKED ISSUE #2225: Binding Watermark to a dynamic property and using UseFloatingWatermark="True" prevents the TextBox from resizing when Text property is empty
See pull request #1510

Dotnetifier:
Fixes #2225

List of changes:
- The `PART_FloatingMessageContainer` grid is now a `ContentControl` because the layout system for ContentControl is way simpler and therefore faster. Also, the grid wasn't necessary.
- The triggers that active the hide/show of the floating watermark are no longer three separate `DataTrigger`s. This system now consists of a `MultiDataTrigger` with two conditions, which both must be satisfied:
  - Whether the input box should have a floating watermark
  - Whether the input box has text
  - The third one, whether the input box actually has a value for `TextBoxHelper.Watermark` has been dropped, because there are no 'negative' DataTriggers. This brings me to the next change:
- The animated property is no longer `Height`; it is now `MaxHeight`. If the watermark is not set, the TextBlock will not take any space because it uses an added `MetroAutoCollapsingTextBlock` style, which is based on `MetroTextBlock` and will automatically collapse when the text property is empty.

### DIFF
--- a/MahApps.Metro/Styles/Controls.ComboBox.xaml
+++ b/MahApps.Metro/Styles/Controls.ComboBox.xaml
@@ -87,10 +87,10 @@
                                        HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}"
                                        VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
                                        Margin="4,2,2,2" />
-                            <Grid x:Name="PART_FloatingMessageContainer"
+                            <ContentControl x:Name="PART_FloatingMessageContainer"
                                   Grid.Column="0"
                                   Grid.ColumnSpan="2"
-                                  Height="0"
+                                  MaxHeight="0"
                                   IsHitTestVisible="False"
                                   Margin="5,0"
                                   Visibility="Visible">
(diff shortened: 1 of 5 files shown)


punker76:
@Dotnetifier thx!

Dotnetifier:
That's fine. I already found a way to solve #2235 when I created this PR, but I'm too much of a git noob to do too many branches and stuff at once :unamused:

punker76:
@Dotnetifier i solved it while testing your changes :-D

Dotnetifier:
Doesn't that make `PART_WatermarkMessage` at [471](https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Styles/Controls.ComboBox.xaml#L471) redundant, if the (floating) watermark responsibility is passed on to `PART_EditableTextBox` at [451](https://github.com/MahApps/MahApps.Metro/blob/master/MahApps.Metro/Styles/Controls.ComboBox.xaml#L451)?

punker76:
@Dotnetifier `PART_WatermarkMessage` is for non editing combobox

Dotnetifier:
@punker76 Ah I see, thanks

PR #2256, opened 2015-12-08
Adding IsCancel="True" to Flyout button

BlythMeister:
This enables a keyboard shortcut to press the escape button to close flyouts


BlythMeister:
Discovered subsequent bug....

PR #2264, opened 2015-12-10
Make combo style respect textbox helper button content

ButchersBoy:
Cant override the button content for combo box; this fixes that.


punker76:
@ButchersBoy thx mate!

PR #2266, opened 2015-12-11
Fix Window Title Thumb Drag

ButchersBoy:
I've got a 3 monitor setup, all different resolutions and orientations.  Native.GetCursorPos reports .y as 200+ something on one monitor, even though the Window is maximised, and thus you can't drag-to-restore.  This seems to be a more robust solution.


punker76:
@ButchersBoy :+1: :ice_cream:

PR #2267, opened 2015-12-11
MVVM dialog unregister instructions. Fixes #2241

### LINKED ISSUE #2241: DialogParticipation.Register holds references to views
We using MahApps + ReactiveUI combo and we have many views which are constantly disposing. If we using DialogParticipation.Register then it holds hard references to registered views and GC can't free them. Perhaps you can use weak references inside DialogParticipation class to avoid this issue.

ButchersBoy:
...just some notes in the XAML to stop people falling into a memory leak trap.


punker76:
@ButchersBoy yeah, documentation 8-D

PR #2307, opened 2016-01-10
Fix TextBoxHelper.SelectAllOnFocus="True"

sachokFoX:
After 3fba68e5 selection on focus stopped working. Add examples to demo application.


punker76:
@sachokFoX :+1:

Issue #2305, opened 2016-01-08
Add a "Remember ..." checkbox to the LoginDialog


manekovskiy:
It would be nice to have a checkbox on `LoginDialog`. From my experience there are a lot of cases when users want to save their username or both username and password.

![checkbox-on-login-dialog](https://cloud.githubusercontent.com/assets/873602/12203721/220bb97e-b639-11e5-9298-9fe6c4d25ff1.png)

The checkbox text and visibility should be configurable in the same way as for example `AffirmativeButton` and `NegativeButton`. And the `LoginDialogData` will be extended with a new flag property indicating whether checkbox is checked or not.

I'm definitely going to implement this feature in my fork so if you are interested I'll send a pull request once it is ready.

bigworld12:
i agree this should be added

xxMUROxx:
Agree :+1:

But then we have to store some information into the application settings. Each *logon dialog* must have an ID that identifies the dialogue, whether it should be opened or not---depending on the selection.

manekovskiy:
@xxMUROxx When we need to access anything that requires authentication we typically have some abstractions like "Authenticator" and "CredentialsStore". Authenticator can either ask CredentialsStore for saved credentials or start the authentication process from the very beginning which involves collecting a username/password using the LoginDialog.
And CredentialsStore knows how to save/retrieve credentials, it is a way to abstract out the place where credentials live between user sessions.

When I'm trying to imagine interactions between this Authenticator/CredentialsStore pair I cannot see "credentials identifier" as a part of LoginDialog, it would be more practical to have it as a part of Authenticator implementation since he is the guy who talks with CredentialsStore.

thoemmi:
@xxMUROxx why should the settings be stored by MahApps.Metro? Username and password aren't stored either. It's up to the application (or business logic respectively) to both provide the login dialog with a `LoginDialogSettings` instance and to process its returned `LoginDialogData`. MahApps.Metro is meant for the presentation layer only.

PR #2308, opened 2016-01-10
Add 'Remember' CheckBox to the LoginDialog

manekovskiy:
This adds a "Remember ..." checkbox to the `LoginDialog`. Reference feature https://github.com/MahApps/MahApps.Metro/issues/2305.

To keep backward compatibility the `Checkbox` is `Collapsed` by default. 
The default text for `Checkbox` equals to "Remember".


MahApps.Metro/Controls/Dialogs/LoginDialog.cs
```diff
@@ -36,12 +40,17 @@ public LoginDialogSettings()
...
+        public Visibility RememberCheckBoxVisibility { get; set; }
+
+        public string RememberCheckBoxText { get; set; }
     }
 
     public class LoginDialogData
     {
         public string Username { get; set; }
         public string Password { get; set; }
+        public bool Remember { get; set; }
```
xxMUROxx: I would change the name of this property to a more `boolean-sounding` name. I would suggest something like *IsRememberRequested*, *RequestRemember*. This is because the developer might be confused because it might understand that MahApps does remember the state of the dialog.
Another possibility would be to add a documentation tag to the property.

thoemmi: I'd suggest to make **all** setters of `LoginDialogData` `internal`. There's no need to set these properties outside of MahApps.Metro.

manekovskiy: Yes, I was also thinking about `ShouldRemember` as alternative to `Remember`. It is more explicit.
Totally agree on suggested `internal` improvement.

PR #2322, opened 2016-01-14
[RFC] Hotkey control

thoemmi:
This PR adds a `HotKeyBox` to MahApps.Metro with following features:
* handles all hotkeys except <kbd>Tab</kbd>, <kbd>Shift+Tab</kbd> (used for navigation) and <kbd>Back</kbd> (clears the hotkey)
* optionally modifier keys (<kbd>Ctrl</kbd>, <kbd>Alt</kbd>, <kbd>Shift</kbd>, or <kbd>Win</kbd>) are required
* supports watermark
* validation works
* templating support
* uses Win API [`GetKeyNameText`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646300%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) to get a textual representation of the keys. No localization is required, as it only depends on the keyboard layout as configured in Windows. (I have a German keyboard, therefore you see `STRG+UMSCHALT+POS1` in the screenshot; with an English layout it would say `CTRL+SHIFT+HOME`)
* supports `ControlsHelper.ContentCharacterCasing`. The reason was that Windows is not consistent about what `GetKeyNameText` returns. E.g. on my machine with an English keyboard layout it returns normal-cased modifier keys, but upper-case function keys, which results in `Ctrl+Shift+HOME`. You may attach `Controls:ControlsHelper.ContentCharacterCasing="Upper"` to get consistent output.

image

**The only purpose of `HotKeyBox` is to enable the user to *enter* a hotkey. Registering and handling hotkeys is out of its scope.** I have added the control to the sample app and use [**NHotkey**](https://github.com/thomaslevesque/NHotkey) ([NuGet](https://www.nuget.org/packages/NHotkey.Wpf/)) for the handling.

### DIFF
--- a/MahApps.Metro/Controls/HotKeyBox.cs
+++ b/MahApps.Metro/Controls/HotKeyBox.cs
@@ -8,7 +8,8 @@ using MahApps.Metro.Native;
 
 namespace MahApps.Metro.Controls {
     [TemplatePart(Name = PART_TextBox, Type = typeof(TextBox))]
-    public class HotKeyBox : Control {
+    public class HotKeyBox : Control
+    {
         private const string PART_TextBox = "PART_TextBox";
 
         public static readonly DependencyProperty HotKeyProperty = DependencyProperty.Register(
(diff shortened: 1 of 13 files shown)


bigworld12:
exactly what I have been looking for, thanks!

punker76:
@thoemmi :+1:

moser96:
Create work :+1: .
I just included it in my project and it's working as expected ;)

thoemmi:
Thanks for your feedback, @moser96. Glad to hear that :smile:

PR #2326, opened 2016-01-20
DragMove fix for #2324

### LINKED ISSUE #2287: IsWindowDraggable broke starting in 1.2
Is this broke in 1.2 or did the behavior change?  I don't appear to be able to drag even a basic vanilla window by mouse clicking/dragging on any position except the title bar.

### LINKED ISSUE #2288: First 30 units of flyout aren't clickable
Hey,
I just updated to the new MahApps.Metro version and noticed the the first 30 units of a flyout aren't clickable because there lays an invisible grid.
FlyoutNotClickable
(As you can see on the picture, even if the mouse is directly over the button, the hover style doesn't get applied)

First, I was sure that it was my fault and I forgot to hide a grid, so I wrote some code which shows the element directly under the cursor in the title bar.
Small Video

Here is the code to display the current element in the title bar (you can also set a break point to analyze the element):
```
        private void MainWindow_OnMouseMove(object sender, MouseEventArgs e)
        {
            var element = Mouse.DirectlyOver as UIElement;
            Title = element?.ToString() ?? "";
        }
```

I will also take a look at the mahapps code but I thought that it would be the best if I create an issue first, perhaps somebody knows where the grid comes from.

### LINKED ISSUE #2324: DataGrid activated by Titlebar
Hello,

I've found a weird issue where the DataGrid is activated when clicking the Titlebar of separate windows within the same application.

Requirements:
* 2 WPF windows with any MahApps theme.
- The first window containing only a DataGrid with say 100 items and 6 columns (each subitem also containing items).
- The second window containing nothing (but can contain anything).

The bug becomes very, very obvious under the following conditions:
1. The first window is maximized.
2. The second window is positioned somewhere on the left of the screen and must be anywhere between columns 1-4.

Sample project can be downloaded here:
http://1drv.ms/23c9XLg

Repro Steps:
Run the sample project, click the "Open Secondary Window" button at the top right then simply click and drag the secondary window to see items on the DataGrid of the first window becoming highlighted.

I've tested this on Windows 7 and Windows 10 from MahApps versions 1.1.2 to 1.2.3 and confirmed the behavior is still present but if you make the main window smaller it becomes much harder to reproduce and it also helps to position the secondary window Titlebar closer to the Datagrid headers of the first window.

If anyone can confirm the bug or knows a workaround it would be appreciated :)

punker76:
- IsHitTestVisibleInChrome for PART_WindowTitleThumb
- use some internal code in DoWindowTitleThumbMoveOnDragDelta from the org DragMove solves the mouse hover on DataGrid and ListBox etc. but not if the second window is maximized. it seems that the mouse capture is lost on the window...
- close #2288 First 30 units of flyout aren't clickable
- close #2324 DataGrid activated by Titlebar
- close #2287 IsWindowDraggable broke starting in 1.2 #2287

2016-01-20 17h18_29


punker76:
I've got it! :see_no_evil:

punker76:
2016-01-21 11h55_52

PR #2335, opened 2016-02-01
Added SelectorDialog

BlythMeister:
This is a dialog box with a selector input.

Example:

image

when selector expanded:

image

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/DialogManager.cs
+++ b/MahApps.Metro/Controls/Dialogs/DialogManager.cs
@@ -139,6 +139,68 @@ namespace MahApps.Metro.Controls.Dialogs
                     }).Unwrap().Unwrap();
                 }));
             }).Unwrap();
+        }/// <summary>
+         /// Creates a Selector inside of the current window.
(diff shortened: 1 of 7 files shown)


thoemmi:
I honor your work, but I'm against adding this dialog to MahApps.Metro. IMHO you cannot add each and every dialog permutation to the library. What if one user requests a dialog with one textbox and one combobox? Or two comboboxes? What about autocompletion? Maybe one user wants a filepicker, i.e. an editable combobox with file path completion? 

No offense, @BlythMeister, but I think a line must be drawn somewhere. However, that's just my $&nbsp;0.02 :neutral_face:

BlythMeister:
I actually agree.
I thought adding a custom dialog for this would be hard, and in hindsight, was pretty easy, even to get they styling.

Going to close :)

thoemmi:
Since you're experienced now with implementing and styling your own custom dialog: how about writing an article about it? I guess @punker76 would appreciate it and add it to the [guides](http://mahapps.com/guides/).

PR #2359, opened 2016-02-18
Added ContentStringFormat TemplateBinding to ToolTip

nrpog:
Fixed a bug in the ToolTip style where the ContentStringFormat was being ignored.


punker76:
@nrpog :+1:

Issue #2370, opened 2016-02-24
Pasting over selection in NumericUpDown doesn't work


michaelprewecki:
In the NumericUpDown control when you attempt to paste a decimal over an existing decimal value nothing happens. 

I would expect a clipboard value containing a decimal value to replace the existing decimal value.

--
### Environment

- MahApps.Metro v1.2.4.0
- Windows 10
- Visual Studio 2013
- .NET Framework 4.5

PR #2371, opened 2016-02-24
Paste over selection for a NumericUpDown control

michaelprewecki:
When pasting into a NumericUpDown control any selected text is now replaced with the pasted text.

Closes Issue #2370 

Fixed up the pasting into the NumericUpDown control so that selected text is replaced.


punker76:
@michaelprewecki :+1:

Issue #2148, opened 2015-09-29
MetroProgressBar orientation does not work


widgetqueen:
Good afternoon!

The Orientation field on the MetroProgressBar control does not seem to work - I tried the workaround suggested here: https://github.com/MahApps/MahApps.Metro/issues/1465 but it does not work for my purposes. Rotating the progress bar creates issues within my grid/viewbox structure that I don't have when simply using a vertically oriented progress bar.

punker76:
@widgetqueen what should be happen if `IsIndeterminate="True"`? should it be come from down to up o from up to down?

widgetqueen:
I'm not sure I understand your question. When I set IsIndeterminate="True", the progress bar disappears. Currently, when I set IsIndeterminate to false and the orientation to vertical, the progress bar does not fill in correctly according to its value. The progress bar itself appears to be vertical, but when the fill is applied, it still appears to fill in horizontally.

petvetbr:
I'm taking a look at this issue.

realies:
This is still not working on the latest alpha. Current solution:

```
<Style TargetType="Controls:MetroProgressBar">
 <Setter Property="LayoutTransform">
  <Setter.Value>
   <RotateTransform CenterX="0.5" CenterY="0.5" Angle="-90"/>
  </Setter.Value>
 </Setter>
</Style>
```

punker76:
@widgetqueen @petvetbr fixed with d4aa412870a9ae0898a39813a71aad87f572c560

PR #2379, opened 2016-02-28
Ajustment for vertical progressbar to work

petvetbr:
## Ajustment for vertical progressbar to work

Ajustment of the MetroProgressBar control template XAML, so that it also works when Orientation=Vertical 

#2148


punker76:
@petvetbr thx! added in d4aa412870a9ae0898a39813a71aad87f572c560

Issue #2166, opened 2015-10-16
Controls Do not get distinct automation ID's


jublin:
When using controls such as NumericUpDown, I am unable to get a unique automationID from the control. They all show up as this: The name is being set in the Xaml. Unfortunately I cannot really give a code snippet.
![mahappsautiomationid](https://cloud.githubusercontent.com/assets/1799126/10550290/ff9cdd08-73fa-11e5-9a29-f81ab6103251.png)

The problem is my group is trying to test a window with multiple NumericUpDowns and other controls with this same issue. Automating a test becomes impossible to do with this.

punker76:
@jublin it seems you should give your controls a `x:Name="xyz123"` for the automatic tester...

jublin:
@punker76 I have done that. 

```
<metroControls:NumericUpDown StringFormat="N0" 
 DataContext="{Binding Source=(Removed)" 
HorizontalAlignment="Left" HorizontalContentAlignment="Left" 
KeyDown="Offset_KeyDown" 
Value="{Binding Offset}" Width="100" x:Name="OffsetControl"  
/>
```

With Name being defined as x:Name= or Name=, I get 

![image](https://cloud.githubusercontent.com/assets/1799126/10704127/7dd4952e-7989-11e5-9170-aea605d8a310.png) 

The updown buttons also have no name. The overall control does not appear as a focus element.

punker76:
@jublin what automatic tool is this?

jublin:
@punker76 This is the "VisualUIAVerify" tool from the windows SDK

punker76:
@jublin ok, thx, i will take a look after releasing v1.2.0

jublin:
This also applies to the Back/down buttons on MetroWindow Flyouts. The automation ID is seen as "nav" for each one. Getting a back button by automationID in testing may not return the intended button for the flyout.

petvetbr:
@punker76 I'm working on a solution for this

petvetbr:
I´m working on a new implementation based on AutomationPeers. Should be available soon.

lightwalker01:
@punker76, @petvetbr any news on this issue?
[...]

PR #2381, opened 2016-02-29
AutomationId based on name fix

petvetbr:
## Set AutomationId based on x:Name property

Now the AutomationId property is set according to the control x:Name property.

For NumericUpDown:

- Textbox: "PART_TextBox{x:Name value}"
- DownButton : "PART_NumericDown{x:Name value}"
- UpButton: "PART_NumericUp{x:Name value}"

For Flyout:

-NavButton: "nav{x:Name value}"

#2166

### DIFF
--- a/MahApps.Metro/Themes/Flyout.xaml
+++ b/MahApps.Metro/Themes/Flyout.xaml
@@ -20,6 +20,7 @@
                     Height="40"
                     VerticalAlignment="Bottom"
                     Style="{DynamicResource MetroCircleButtonStyle}"
+                     AutomationProperties.AutomationId="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Name,  StringFormat='nav{0}' , Mode=OneWay }"
                     Foreground="{Binding RelativeSource={RelativeSource AncestorType={x:Type Controls:Flyout}}, Path=Foreground}"
                     FontFamily="Segoe UI Symbol"
                     FontSize="16"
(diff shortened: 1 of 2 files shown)


punker76:
@thoemmi What do you think? /cc @xxMUROxx

thoemmi:
I don't know anything about the automation API, but it looks fine for me :blush:

xxMUROxx:
Can someone explain in short what **Automation** is useful for?

petvetbr:
@xxMUROxx there are 2 areas where automation is used a lot:

Automated UI tests
Accessibility

In both it is important to manipulate the UI using alternatives to de usual keyboard/mouse/touch methods, so that is when automation comes into play.

xxMUROxx:
@petvetbr Ah, ok. Then I think the same behavior is missing for other custom controls that are implemented in this library too. e.g. ToggleSwith, DropDownButton, SplitButton, etc...

petvetbr:
@xxMUROxx

Ok, I´ll take a look at the other controls and see if I can make them more "automation friendly",

xxMUROxx:
@petvetbr :+1: 

One thing to point out. For custom controls created by this library you could introduce a method in e.g. NumericUpDown that attached the properties to the PART_ controls in order to avoid that the attached property is called in XAML. This has the advantage that if someone rewrites the template for e.g. NumericUpDown the _automation_ is applied too.

private void ApplyUiAutomation()
{
    _repeatDown.SetValue(Controls:AutomationHelper.DefaultAutomationId, ElementNumericDown);
    //...
}
what do you think?

/cc @punker76

xxMUROxx:
I have looked into it a bit and found that the UI Elements should get exposed by [AutomationPeer Class](https://msdn.microsoft.com/en-us/library/system.windows.automation.peers.automationpeer.aspx). So it is done with ButtonAutomationPeer, TextBoxAutomationPeer.

Therefore I think it should be done also with NumericUpDown. Create a class NumericUpDownAutomationPeer and implement the exposure there

[UI Automation of a WPF Custom Control](https://msdn.microsoft.com/library/cc165614(v=vs.100).aspx)

Issue #2204, opened 2015-11-09
Progress ring dot scaling seems to be wrong (1.1.2)?


dsaf:
Microsoft [UX guidelines](https://msdn.microsoft.com/en-gb/library/windows/apps/hh465469.aspx):

![image](https://cloud.githubusercontent.com/assets/3845650/11036166/b6e480dc-86ee-11e5-9389-58806ed8e6a2.png)

MahApps.Metro Demo App:
![image](https://cloud.githubusercontent.com/assets/3845650/11036179/cf62e40a-86ee-11e5-810e-245b5cb41b14.png)

If you compare them, the standard version has five dots instead of six and the dots are relatively larger. The number of dots is configurable via the ```IsLarge``` property, but the size of dots ```EllipseDiameter``` is unfortunately readonly.

![image](https://cloud.githubusercontent.com/assets/3845650/11036764/4cbae936-86f2-11e5-8424-564622dcd420.png)

dsaf:
Thank you.

PR #2382, opened 2016-02-29
Added EllipseDiameterScale (issue #2204)

petvetbr:
## Added EllipseDiameterScale (issue #2204)

Added EllipseDiameterScale property as a multiplier for Ellipse Size, below the scaling of 1.0 (default), 0.5 (smaller) and 1.5 larger values for this property are shown.

This solves issue #2204 as now it is possible do manually ajust the size of the Ellipse. In the control the Ellipse size is calculated using the formula below:

EllipseDiameter =(width / 8)*EllipseDiameterScale;

ring0_5
ring1
ring1_5

### DIFF
--- a/MahApps.Metro/Controls/ProgressRing.cs
+++ b/MahApps.Metro/Controls/ProgressRing.cs
@@ -25,6 +25,8 @@ namespace MahApps.Metro.Controls
 
         public static readonly DependencyProperty EllipseOffsetProperty = DependencyProperty.Register("EllipseOffset", typeof(Thickness), typeof(ProgressRing), new PropertyMetadata(default(Thickness)));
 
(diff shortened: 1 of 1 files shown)


punker76:
@petvetbr Nice to set the ellipse size, but it solves not the issue "... the standard version has five dots instead of six and..."

petvetbr:
@punker76 The control will show 5 ellipse when it is set to the IsLarge=false property, otherwise it shows 6 ellipses as @dsaf mentioned in the original comment:

> The number of dots is configurable via the IsLarge property, but the size of dots EllipseDiameter is unfortunately readonly.

punker76:
@petvetbr ah, sorry, you're right :see_no_evil:

punker76:
@petvetbr :+1:

petvetbr:
@punker76 Glad to help. I´m with some spare time this week, so if there is any issue that you want me to look, just ask.

Issue #2380, opened 2016-02-29
Missing FontSize Property for ProgressDialog/ShowProgressAsync


Memphizzz:
## What steps will reproduce this issue?

Unable to set FontSize for ProgressDialog.

### Expected outcome

Being able to set FontSize either using ShowProgressAsync parameter or using MetroDialogSettings.

--
### Environment

- MahApps.Metro __v1.2.4__
- Windows __10__
- Visual Studio __2015__
- .NET Framework __4.5__

PR #2383, opened 2016-02-29
Dialog Font Sizes Setting Implementation #2380

petvetbr:
## Implementation of the option to set Dialog Font sizes using MetroDialogSettings object.

The size of the fonts for the Title and Message of a Dialog, can be set using a MetroDialogSettings object and passing it as a parameter.

Example:

`var controller = await this.ShowProgressAsync("Please wait...", "We are baking some cupcakes!", false, new MetroDialogSettings() { DialogMessageFontSize = 25, DialogTitleFontSize = 50 });`

 #2380

### DIFF
--- a/MahApps.Metro/Styles/Fonts.xaml
+++ b/MahApps.Metro/Styles/Fonts.xaml
@@ -25,9 +25,6 @@
 
     <System:Double x:Key="StatusBarFontSize">12</System:Double>
 
-    <System:Double x:Key="DialogTitleFontSize">26</System:Double>
-    <System:Double x:Key="DialogMessageFontSize">15</System:Double>
-
     <System:Double x:Key="FlyoutHeaderFontSize">20</System:Double>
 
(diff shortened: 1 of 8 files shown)


MahApps.Metro/Styles/Fonts.xaml
```diff
@@ -25,9 +25,6 @@
 
     <System:Double x:Key="StatusBarFontSize">12</System:Double>
 
-    <System:Double x:Key="DialogTitleFontSize">26</System:Double>
-    <System:Double x:Key="DialogMessageFontSize">15</System:Double>
-
     <System:Double x:Key="FlyoutHeaderFontSize">20</System:Double>
```
punker76: Can you try reuse this existing font sizes? Otherwise it's a breaking change and I can't add this PR for 1.3.0

petvetbr: @punker76 sure, I´ll do that.

PR #2386, opened 2016-03-01
Automation improvements #2381

petvetbr:
## Improved automation support with AutomationId and Name automatically set according with control Name property.

Following @xxMUROxx sugestion on issue #2381:

Implemented an AutomationHelper, so that the AutomationId and Name are setted based on the Name of the Control for most of the custom controls components. The goal is to eventually obtain full automation coverage for UI Automated Testing and Accessibility purposes.

#2381


punker76:
@petvetbr you don't use the latest source from develop branch

xxMUROxx:
I have looked into it a bit and found that the UI Elements should get exposed by [AutomationPeer Class](https://msdn.microsoft.com/en-us/library/system.windows.automation.peers.automationpeer.aspx). So it is done with ButtonAutomationPeer, TextBoxAutomationPeer.

Therefore I think it should be done also with NumericUpDown. Create a class NumericUpDownAutomationPeer and implement the exposure there

[UI Automation of a WPF Custom Control](https://msdn.microsoft.com/library/cc165614(v=vs.100).aspx)

xxMUROxx:
I have tried it with DateTimePicker of MahApps

~~~
public class DateTimePickerAutomationPeer : FrameworkElementAutomationPeer
    {
        public DateTimePickerAutomationPeer(DateTimePicker owner)
            : base(owner)
        {
        }

        private DateTimePicker OwningDateTimePicker
        {
            get { return Owner as DateTimePicker; }
        }

        protected override List<AutomationPeer> GetChildrenCore()
        {
            List<AutomationPeer> peers = new List<AutomationPeer>();
            var peer = CreatePeerForElement(OwningDateTimePicker._button);
            if (peer != null)
            {
                peers.Add(peer);
            }
            peer = CreatePeerForElement(OwningDateTimePicker.SecondInput);
            if (peer != null)
            {
                peers.Add(peer);
            }
            return base.GetChildrenCore();
        }
    }
~~~

To work the controls like **_secondInput** in **DateTimePicker** must be exposed as internal property **SecondInput**.

petvetbr:
@xxMUROxx Ok. I'll start implementing automation peers for the controls. The first solution I proposed was faster to implement, but you are right, we should try to follow the same patterns recomended by MS and when ready I'll submit a new PR.

punker76:
@petvetbr I close this and will wait for a new PR, thx

Issue #2189, opened 2015-10-30
Option to make NumericUpDown empty until value entered


jacbz:
By default, the NumericUpDown will always display a value. I'm mostly using the StringFormat "N3", which adds thousand separators and three decimal places. When entering a value, the default "0.000" is more a hindrance than anything; you'd have to delete the text before entering a value.

Is there a way to set an empty default text? I've tried setting the StringFormat to "", but that works only for the first time, after the second time the default value becomes 0 again.

punker76:
Long time ago... ok. Yes, I think this could be possible using a nullable type and the `Controls:TextBoxHelper.Watermark` property. So, if the value is `null` the watermark will be displayed.

![2018-01-18_00h12_54](https://user-images.githubusercontent.com/658431/35072327-6079d572-fbe4-11e7-8c78-b0bcb77c487e.png)
![2018-01-18_00h13_03](https://user-images.githubusercontent.com/658431/35072328-60983d14-fbe4-11e7-98e9-6b781b626a87.png)

PR #2391, opened 2016-03-02
NumericUpDown text option for zero or null #2189

petvetbr:
## What changed?

Created TextIfZero and TextIfNull properties:

**TextIfZero**: If set to any other value than null, the property value will be shown when the NumericUpDownValue is equal to zero.

textzero

**TextIfNull**: If set to any other value than null, the property value will be shown when the NumericUpDownValue is null.

textnull
#2189


xxMUROxx:
Well the idea is good :+1:

thoemmi:
I'm not fond of this :-/

xxMUROxx:
According to #2189 a compromise would be to show in that case only **0** instead of the complete `StringFormat` (e.g. **0.000**)
For the other case, now I agree with @thoemmi.

Issue #2333, opened 2016-01-30
FlipView SelectedIndex resetting bug


berkay2578:
Update:
--
After a bit of tinkering I found out that whether FlipView.SelectedIndex is attached to a Binding, changing the active Tab will make it go back to Index 0. However, this will only happen once. If you change the Tab and then go back, FlipView will no longer reset back to Index 0 after you change tabs.

-----

I'm experiencing this bug in my own project, which you can access through https://github.com/berkay2578/nfsw-server. You can find the code-behind related to the FlipView in the MainWindow.xaml.cs on the lines 104-131; or:
```
#region FlipViewPersona
            FlipViewPersonaImage.HideControlButtons();

            Grid[] aFlipViewAvatarArray = new Grid[28];
            for (int i = 0; i < 28; i++)
            {
                Grid Grid_FlipViewDummy;
                Image Image_FlipViewDummy;
                Image_FlipViewDummy = new Image() { Margin = new Thickness(7d), HorizontalAlignment = HorizontalAlignment.Center, VerticalAlignment = VerticalAlignment.Center, Stretch = Stretch.Uniform, Source = (ImageSource)BitmapFrame.Create(new Uri("pack://application:,,,/OfflineServer;component/images/NFSW_Avatars/Avatar_" + i.ToString() + ".png", UriKind.Absolute)) };
                Grid_FlipViewDummy = new Grid() { Margin = new Thickness(-8d) };
                Grid_FlipViewDummy.Children.Add(Image_FlipViewDummy);
                Image t1 = new Image() { Source = Image_FlipViewDummy.Source };
                t1.Effect = new System.Windows.Media.Effects.BlurEffect() { Radius = 7.5d };
                Grid_FlipViewDummy.Background = new VisualBrush((Visual)t1);
                aFlipViewAvatarArray[i] = Grid_FlipViewDummy;
            }
            FlipViewPersonaImage.ItemsSource = aFlipViewAvatarArray;

            Binding indexBind = new Binding()
            {
                Path = new PropertyPath("mPersona.shAvatarIndex"),
                UpdateSourceTrigger = UpdateSourceTrigger.PropertyChanged,
                Mode = BindingMode.TwoWay,
                Source = mCurrentSession
            };
            FlipViewPersonaImage.SelectedIndex = mCurrentSession.mPersona.shAvatarIndex;
            BindingOperations.SetBinding(FlipViewPersonaImage, FlipView.SelectedIndexProperty, indexBind);
[...]

PR #2392, opened 2016-03-02
Correction of lost SelectionIndex #2333

petvetbr:
## What changed?

I got the same behavior described by @berkay2578 using the example he published on https://github.com/berkay2578/nfsw-server

It seems that the problem was related to bootstraping controls events and setting the SelecionIndex=0 in the Loaded event handler and then unsubscribing everything on the Unloaded event.

As described in the control comments and also in this stackoverflow question (http://stackoverflow.com/questions/6349695/removing-event-handlers-from-custom-controls-template-parts), when the control is inside a TabItem it can be loaded and unloaded multiple times, so everytime the control was unloaded and then loaded again, for example when the user changed tabs, it resetted the SelecionIndex to zero.

Now the event bootstrapping and initial setting of SelectionIndex occurs inside the OnApplyTemplate event. 

#2333

### DIFF
--- a/MahApps.Metro/Controls/FlipView.cs
+++ b/MahApps.Metro/Controls/FlipView.cs
@@ -57,8 +57,6 @@ namespace MahApps.Metro.Controls
         
         public FlipView()
         {
-            this.Unloaded += FlipView_Unloaded;
-            this.Loaded += FlipView_Loaded;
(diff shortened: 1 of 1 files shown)


punker76:
Fixed by 2ffa52cddb8da4c3d2485495d3cd8d4e2c1760a9

Issue #1835, opened 2015-03-11
Window animation interferes with WindowsFormsHost positioning


barbeque:
When you put a WindowsFormsHost inside a Metro window, sometimes the WinForms control will occasionally 'hang' in space after the introductory animation has completed and only move once an event (such as hovering over the 'close' box or tabbing out of the `MaskedTextBox`) has been processed.

![image](https://cloud.githubusercontent.com/assets/423811/6607409/267bbb0e-c804-11e4-85b0-864833f40edd.png)

The actual control in question doesn't seem to matter; I've experienced it with `Button`, `MaskedTextBox` and `CheckedListBox` off the top of my head. It also happens if the `WindowsFormsHost` is located inside of a `StackPanel` or even the root Content control, not just the nested `GroupBox` described in the example XAML.

The test XAML to produce this window is:

``` xaml
    <metro:MetroWindow x:Class="MetroDemo.DemoWindow2"
            xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
            xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
            xmlns:metro="clr-namespace:MahApps.Metro.Controls;assembly=MahApps.Metro"
            xmlns:wfh="clr-namespace:System.Windows.Forms;assembly=System.Windows.Forms"
            Title="Demo window 2" Height="500" Width="400">
        <DockPanel LastChildFill="True">
            <StackPanel Orientation="Vertical" DockPanel.Dock="Bottom">
                <GroupBox Header="WinForms Controls">
                    <StackPanel Orientation="Vertical">
                        <WindowsFormsHost Width="100" HorizontalAlignment="Left">
                            <wfh:MaskedTextBox Mask="00/000/0000"/>
                        </WindowsFormsHost>
                    </StackPanel>
                </GroupBox>
            </StackPanel>
        </DockPanel>
    </metro:MetroWindow>
```

I believe it is animation related because setting `WindowTransitionsEnabled` to false fixes the problem.

I am using v1.1.2.0, grabbed from NuGet yesterday. OS is Windows 8.1 x64.

kjeremy:
I see this too on Windows 7

punker76:
[...]

PR #2394, opened 2016-03-02
Issue Winhost "hanging" after transition #1835

petvetbr:
## What changed?

I found a StackOverflow post (https://stackoverflow.com/questions/12465357/force-redraw-after-storyboard-with-windowsformshost-wpf/12748181#12748181) with the same problem and implemented a fix using de solution provided (calling InvalidateVisual() when the transition is completed to force a redraw of the Winform Host).

I believe a more elegant solution would be to have events for when the transition animation is starting or completing so that the developer can decide if it is better to hide the Winform host until the transition is completed, avoiding that the user seeing the winform control abrupty moving to another position when the transition is completed. 

These new events would have to be created in MetroContentControl and MetroWindow controls.
I can do this, but  I´d like to ask if this is something the project maintainers think is a good idea.
#1835

### DIFF
--- a/MahApps.Metro/Controls/MetroContentControl.cs
+++ b/MahApps.Metro/Controls/MetroContentControl.cs
@@ -1,5 +1,7 @@
-﻿using System.Windows;
+﻿using System;
+using System.Windows;
 using System.Windows.Controls;
+using System.Windows.Media.Animation;
 
 namespace MahApps.Metro.Controls
 {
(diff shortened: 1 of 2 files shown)


punker76:
@petvetbr Thx for this. I changed it a little bit, I added a `TransitionCompleted` event to `MetroContentControl` and `WindowTransitionCompleted` event to `MetroWindow` which will be fired after the loaded Storyboard is completed.

PR #2423, opened 2016-03-10
Added the habilitiy to block the use of the scientific notation 'e'.

### LINKED ISSUE #2422: Added the habilitiy to block the use of the scientific notation 'e' on NumericUpDown.
This is not an issue, but a feature request.
I would like to block the user from using the scientific notation 'e' inside the `NumericUpDown`, so only numbers would be allowed.

I took the liberty to code it myself, so the PR is coming soon.
## 
### Environment
- MahApps.Metro **dev branch**
- Windows **7x64**
- Visual Studio **2015**
- .NET Framework **4.5**

feinstein:
## What changed?

Added the habilitiy to block the use of the scientific notation 'e', using the property HasScientificNotation. Updated the example as well.

closes #2422.


punker76:
1  @xxMUROxx @thoemmi ?

xxMUROxx:
I would rename the DP from **HasScientificNotation** to a name that is more saying what the property does. For my understanding the property should---if set to false---remove all `e` values.
What the DP acually does is "allow/disallow the user to press e".

I would suggest to change **HasScientificNotation** to **AllowsScientificNotation**

> The same issue is with **HasDecimals**

But this is a suggestion. Keep free to comment.

MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -463,6 +469,18 @@ public bool HasDecimals
             set { SetValue(HasDecimalsProperty, value); }
         }
 
+        /// <summary>
+        ///     Indicates if the NumericUpDown should show the decimal separator or not.
```
xxMUROxx: The comment should be update to match the correct behavior.

xxMUROxx:
In addition we could introduce a DP that can be set by an enum:

**NumericInputMode** of type **NumericInput**

~~~
[Flags]
public enum NumericInput
{
  NumbersOnly = 1 << 1,
  NumbersDecimal = 2 << 1,
  NumbersHex = 3 << 1,
  All = NumbersDecimal | NumbersHex
}
~~~

That would give us the benefit of adding additional behaviors if requested in the future.

feinstein:
Any decisions on this?

PR #2438, opened 2016-03-21
Remove BindsTwoWayByDefault from IsActiveProperty from ProgressRing

xxMUROxx:
## What changed?

Removed BindsTwoWayByDefault from IsActiveProperty. There is not need to apply TwoWay by default, because the GUI does not update the IsActive property in any parts of the code

# REMARKS

The only change is when **Visibility** changed.

In that case. Would it be important that when the visibility of ProgessRing changes that IsActive is reflected to the **bound property**?

As crossreference, IsIndeterminate has the same behavior
http://referencesource.microsoft.com/#PresentationFramework/src/Framework/System/Windows/Controls/ProgressBar.cs,60ee7d94e2137368,references

### DIFF
--- a/MahApps.Metro/Controls/ProgressRing.cs
+++ b/MahApps.Metro/Controls/ProgressRing.cs
@@ -15,7 +15,7 @@ namespace MahApps.Metro.Controls
     {
         public static readonly DependencyProperty BindableWidthProperty = DependencyProperty.Register("BindableWidth", typeof(double), typeof(ProgressRing), new PropertyMetadata(default(double), BindableWidthCallback));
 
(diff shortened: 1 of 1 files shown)


xxMUROxx:
**Wait:** overseen that is set by static constructor, when `Visibility` changed.

punker76:
@xxMUROxx I don't know if this maybe a breaking functonal break? /cc  @thoemmi

thoemmi:
That is a breaking change. But `ProgressRing.IsActive` never changes on its own, so I don't think anybody expected two-way binding in the past. I'm fine with this PR.

xxMUROxx:
That was my initial thought. I think nobody would use the value of IsActive (set when visibility changes) as part in the backend code. Generally IsActive is only set by backend code to display/hide the contol.

PR #2465, opened 2016-04-11
Virtual Keyboard Service Pack

xxMUROxx:
## What changed?

A new feature is introduced that shows the **[Windows integrated Virtual Keyboard (VK)](https://www.google.it/search?q=windows+virtual+keyboard&source=lnms&tbm=isch)** on devices without an attached keyboard but with existing touch screen.
## How?

Introduced a new Attached Property `VirtualKeyboardHelper.EnableVirtualKeyboard` and added this as setter to the style definition of `TextBox`, `PasswordBox`, `DatePicker`, `DateTimePicker`.

Use `InputScope` on `Controls` to define the default keyboard open style
## TODO
- [x] Fix issue with non opening VK on DatePicker
- [x] Testing: On the fly detaching/attaching of USB Keyboard
- [ ] Testing on Surface Pro 4
  - [ ] Testing: On the fly detaching/attaching of Surface Keyboard
  - [ ] Testing: Fold back of Surface Keyboard
- [ ] Open VK in **handwriting-view** when stylus is present
- [x] ~~Fix issue with too many open keyboard processes~~

### DIFF
--- a/MahApps.Metro/Controls/Helper/VirtualKeyboardHelper.cs
+++ b/MahApps.Metro/Controls/Helper/VirtualKeyboardHelper.cs
@@ -24,7 +24,7 @@ namespace MahApps.Metro.Controls.Helper
             }
             else
             {
-                startInfo = new ProcessStartInfo(@"C:\Program Files (x86)\Common Files\Microsoft Shared\ink\TabTip32.exe");
+                startInfo = new ProcessStartInfo(@"C:\Program Files\Common Files\Microsoft Shared\ink\TabTip32.exe");
             }
             startInfo.WindowStyle = ProcessWindowStyle.Hidden;
 
--- a/MahApps.Metro/Controls/Helper/VirtualKeyboardHelper.cs
+++ b/MahApps.Metro/Controls/Helper/VirtualKeyboardHelper.cs
@@ -1,0 +1,102 @@
+﻿using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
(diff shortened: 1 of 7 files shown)


MahApps.Metro/Controls/Helper/VirtualKeyboardHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+        private static readonly Lazy<ProcessStartInfo> StartInfo = new Lazy<ProcessStartInfo>(() =>
+        {
+            ProcessStartInfo startInfo;
+            if (Environment.Is64BitOperatingSystem)
+            {
+                startInfo = new ProcessStartInfo(@"C:\Program Files\Common Files\Microsoft Shared\ink\TabTip.exe");
+            }
+            else
+            {
+                startInfo = new ProcessStartInfo(@"C:\Program Files (x86)\Common Files\Microsoft Shared\ink\TabTip32.exe");
```
dgadelha: If it isn't a 64-bit version of Windows, `Program Files (x86)` does not exist. This directory is only present in 64-bit versions for Windows, and in 32-bit its called `Program Files`.

xxMUROxx: @dgadelha Thanks

xxMUROxx: @dgadelha do you have a x86 OS? can you check whether the file TabTip32.exe exists?

PR #2467, opened 2016-04-12
Added Canceled Result To MessageDialogResut

MatthewKinsman:
## What changed?

I added a canceled state to MessageDialogResult enum.  I had an issue with cross thread access when canceling the dialog.  It would attempt to check ButtonStyle which is owned by another thread.

### DIFF
--- a/MahApps.Metro/Controls/Dialogs/MessageDialog.cs
+++ b/MahApps.Metro/Controls/Dialogs/MessageDialog.cs
@@ -77,7 +77,7 @@ namespace MahApps.Metro.Controls.Dialogs
             var cancellationTokenRegistration = DialogSettings.CancellationToken.Register(() =>
             {
                 cleanUpHandlers?.Invoke();
-                tcs.TrySetResult(ButtonStyle == MessageDialogStyle.Affirmative ? MessageDialogResult.Affirmative : MessageDialogResult.Negative);
+                tcs.TrySetResult(MessageDialogResult.Canceled);
             });
 
             cleanUpHandlers = () => {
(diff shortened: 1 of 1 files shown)


Deadpikle:
@MatthewKinsman -- You may also consider sending canceled if the user hits the 'Escape' key (see escapeKeyHandler).

thoemmi:
I like the introduction of `Canceled` (and second @Deadpikle's suggestion for the <kbd>Esc</kbd> key). However, this is a **major breaking change**:
Users of MahApps.Metro would have to check each and every usage of MessageBox in their code. Their code still compiles, but the the MessageBox' result is now different. I think this is really dangerous.

My personal sidenote: I'd prefer `Cancelled` to `Canceled`, see [Canceled vs. cancelled](http://grammarist.com/spelling/cancel/). However, even Microsoft is not consistent, there's [`AsyncCompletedEventArgs.Cancelled`](https://msdn.microsoft.com/en-us/library/system.componentmodel.asynccompletedeventargs.cancelled(v=vs.110).aspx) and [`BackgroundTaskInstance.Canceled`](https://msdn.microsoft.com/en-us/library/windows/apps/windows.applicationmodel.background.ibackgroundtaskinstance.canceled.aspx).

punker76:
@MatthewKinsman I'm also with @thoemmi . So this change should go to the dev-2.0.0 branch... or you/we wait until I finished the v1.3.0 which is .... nearly finished :-D

punker76:
The Canceled value is now added without breaking change, cause since #2619 "New Nullable Property MessageDialogResult? DialogResultOnCancel"

Issue #2398, opened 2016-03-04
System.TypeLoadException with SpellCheck.IsEnabledProperty


punker76:
## What steps will reproduce this issue?

from @gigios

>I'm not sure if this is the correct location where post my problem, but it is related the latest release of sources and paket usage.
Probably the problem is related to my configuration, but I'm not sure.

>After one month I have downloaded the new sources and installed the paket extension in the VS2015, open the solution and started the MetroDemo app (FW 4.0). No build error are present but at runtime I receive the exception: An exception of type 'System.TypeLoadException' occurred in mscorlib.dll but was not handled in user code.
The exception is raised in the TextBoxHelper at the line 74 (tb.SetValue(SpellCheck.IsEnabledProperty, true);) of the method UseSpellCheckContextMenuChanged.
If I start the 4.5 version of the demo all works fine.
For my project this is not a problem because I'm not using the SpellCheck context menu, but I'm using the framework 4.0 and I don't know if there are other possible problems.

### Expected outcome

No exception under .NET 4.0 and usage o fSpellCheck.IsEnabledProperty.

--
### Environment

- MahApps.Metro __v1.2.4 (v1.3.0)__
- .NET Framework __4.0__

thoemmi:
I experienced the same exception when running the .NET 4.0 sample app. I'm running the latest Windows 10 Insider Build 14279. The stacktrace looks like this:
```
Could not load type 'System.Attribute' from assembly 'System.Runtime, Version=2.6.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
   at System.ModuleHandle.ResolveType(RuntimeModule module, Int32 typeToken, IntPtr* typeInstArgs, Int32 typeInstCount, IntPtr* methodInstArgs, Int32 methodInstCount, ObjectHandleOnStack type)
   at System.ModuleHandle.ResolveTypeHandleInternal(RuntimeModule module, Int32 typeToken, RuntimeTypeHandle[] typeInstantiationContext, RuntimeTypeHandle[] methodInstantiationContext)
   at System.Reflection.RuntimeModule.ResolveType(Int32 metadataToken, Type[] genericTypeArguments, Type[] genericMethodArguments)
[...]

PR #2489, opened 2016-05-04
Fix System.TypeLoadException with SpellCheck.IsEnabledProperty #2398 in sample app

davidefu:
For fixing System.TypeLoadException with SpellCheck.IsEnabledProperty in Sample app in the app.config you have to explicit specify the max version
<bindingRedirect oldVersion="0.0.0.0-2.6.10.0" newVersion="2.6.10.0" />
i think that the CLR try to load System.Runtime from the 4.0 version but it's redirected to the 2.6.10 causing the issue

hope this helps!
Davide


punker76:
@davidefu i get the same exceptions as without your changes ...
```
Exception thrown: 'System.InvalidCastException' in PresentationFramework.dll
Exception thrown: 'System.PlatformNotSupportedException' in PresentationFramework.dll
```

davidefu:
this is a different exception..
btw if you simply change the config file you must see the correct behavior

please check the attacched sample.
[net40.zip](https://github.com/MahApps/MahApps.Metro/files/251272/net40.zip)
the only difference between the .bug.exe and the regular .exe is in the config file

punker76:
ba98ffc419b1dafb23ab1f72786c868259a7c64e

Issue #2499, opened 2016-05-12
Time picker control


realies:
I can see that there's a DatePicker control that allows for selecting date and time, but was wondering if the time selection element could be brought up in a new control for the intention of selecting time only?

punker76:
^ @xxMUROxx

xxMUROxx:
@punker Well, nice idea. Can you assign me to this task so that I do not forget it?

PR #2510, opened 2016-05-18
TimePicker control support

xxMUROxx:
As requested in #2499


punker76:
@xxMUROxx Is it ready?

xxMUROxx:
@punker76 sorry for the late answer. Yes it is and glad to see that is was already merged.

PR #2518, opened 2016-05-20
Added methods ShowModalInputExternal, ShowModalLoginExternal and ShowModalMessageExternal.

adeniltonbs:
## What changed?

Methods Added: ShowModalInputExternal, ShowModalLoginExternal and ShowModalMessageExternal in the DialogManager, DialogCoordinator and IDialogCoordinator.
Added samples to use these methods in MainWindow.


adeniltonbs:
@punker76 could check it please. This implementation facilitate my work and believe it will be helpful to others.

punker76:
@thoemmi wdyt?

thoemmi:
I as never fond of external windows in MahApps.Metro, because it's not "Metro-ish". Anyway, looking at the diff it looks ok to mee. However, I would rename `GetExternalWindow` to `CreateExternalWindow`, because that's what it's doing. And `GetExternalMessage` is also misleading, as it creates a new window and sets certain properties. Call it `CreateExternalMessageWindow` or something like that.

BTW, I'd have commented the changes directly/inline, but I'm not a contributor... 😉

PR #2556, opened 2016-06-20
Interact with DialogController from Background Thread

xxMUROxx:
## What changed?

It is possible to show and hide dialogs from any thread

Fixes issue with threads as indicated in [Hide all visible Metro Dialogs before showing another one](http://stackoverflow.com/questions/37789431/hide-all-visible-metro-dialogs-before-showing-another-one)


xxMUROxx:
@punker76 is there any bug in this PR, because of the Bug tag?

punker76:
@xxMUROxx I marked it with the Bug tag cause it's a bug if you fon' t use the Dispatcher ;-)

xxMUROxx:
@punker76 Ok, understood 👍

PR #2575, opened 2016-07-08
Close SplitButton popup if it loses mouse capture

### LINKED ISSUE #2574: SplitButton Popup Doesn't Close After Alt+Tab

1) Open demo app
2) Click dropdown arrow button on the split button in the bottom right (the one that says "Worlds")
3) Alt+Tab a few times
4) Now you can't close the Popup without selecting an item in the Popup's list, even when selecting another control.

On a side note, it'd be good if the popup closed on Alt+Tab as well, probably, since it shows over all windows.

buggy bug

### Expected outcome

Expect the popup to close even if the window has lost focus and/or I've come back to it.

--
### Environment

- MahApps.Metro v1.3.0-ALPHA174 (built from commit 1b12418)
- Windows 10
- Visual Studio 2015
- .NET Framework 4.5

--
### Notes

OutsideCapturedElementHandler in SplitButton.cs doesn't seem to be called. That's probably the reason this is broken. :(

Deadpikle:
## What changed?

If the SplitButton loses mouse capture while its Popup is open, then close the popup.

Closes #2574 -- bug where Popup would stay open even when the window lost focus.

### DIFF
--- a/MahApps.Metro/Controls/SplitButton.cs
+++ b/MahApps.Metro/Controls/SplitButton.cs
@@ -353,12 +353,25 @@ namespace MahApps.Metro.Controls
         {
             Mouse.Capture(this, CaptureMode.SubTree);
             Mouse.AddPreviewMouseDownOutsideCapturedElementHandler(this, this.OutsideCapturedElementHandler);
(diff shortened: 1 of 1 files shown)


punker76:
@Deadpikle Thx!!

Issue #2604, opened 2016-08-06
Add new control: TreeListView


Alkalinee:
Hello,
I wrote a control for a personal project, a TreeListView which is a TreeView with columns like the ListView. The design is based on this library. Here you can see my repository with this control: https://github.com/Alkalinee/MetroTreeListView

![TreeListView](http://i.imgur.com/b7QqbAo.png)
Do you want me to create a pull request?

PR #2650, opened 2016-09-14
Add TreeListView

Alkalinee:
As already mentioned in #2604, I was writing a TreeListView for a personal project which is basically a TreeView with columns (like the ListView).
TreeListView

One of the biggest problems with this control is that you have to somehow set the tree space to the first column and not to the whole GridViewRowPresenter. I solved that by modifying the CellTemplate: I wrapped the original template in a ContentPresenter, set the Margin of the ContentPresenter and finally set the ContentPresenter as the new CellTemplate.

It is always the first column which has the expanding button. If the first column is changed, the template will be reset for and the new first column will be modified.

If you want to test that control:

``` xaml
        <controls:TreeListView ChildrenPropertyName="Children" ItemsSource="{Binding Collection}">
            <controls:TreeListView.View>
                <GridView>
                    <!-- Just add your columns -->
                </GridView>
            </controls:TreeListView.View>
        </controls:TreeListView>
```

I should mention the property `ChildrenPropertyName`. I made the `ItemTemplate` property read only so the developer doesn't get confused why that doesn't work. The TreeListView just doesn't support this property, but because WPF handles `HierarchicalDataTemplate`s really well, the `ChildrenPropertyName` just sets the `HierarchicalDataTemplate` to a new template with the `ItemsSource` = a binding with the `ChildrenPropertyName` as path.
There are two modifications possible:
1. Change the `ChildrenPropertyName` to type Binding (instead of string)
2. Remove the `ChildrenPropertyName` property and the user has to set an empty ItemTemplate by himself.
I don't know which would be better, I like it with a simple property.


punker76:
@Alkalinee Sorry for the late response! Finally I looked now in to the source. It's all fine, but this control works only correct with the same Model type especially the childrens. I'm correct?
@thoemmi What do you think? I can add this, but then we must really declare this.

thoemmi:
I've pulled the request into [my own repository](https://github.com/thoemmi/MahApps.Metro/tree/pr2650-TreeListView) and tried using it.

I added the `TreeListView` to the **Selection** view to show a hierarchy of artists and albums:
```xaml
<TabItem Header="TreeListView">
    <Controls:TreeListView ItemsSource="{Binding Artists}" ChildrenPropertyName="Albums">
        <Controls:TreeListView.View>
            <GridView AllowsColumnReorder="True">
                <GridViewColumn Header="Name" DisplayMemberBinding="{Binding Name}" />
                <GridViewColumn Header="Title" DisplayMemberBinding="{Binding Title}" />
                <GridViewColumn Header="Price" DisplayMemberBinding="{Binding Price}" />
            </GridView>
        </Controls:TreeListView.View>
    </Controls:TreeListView>
</TabItem>
```
This is the result:
image

As you can see the control doesn't work well if you don't have uniform items. AT the top level we have `Artist` with a `Name` property, and at the second level `Album` with `Title` and `Price` (don't ask me why the prices are wrong; I don't know)

The `TreeListView` seems to work well if you have uniform, hierarchical data. Good work, @Alkalinee. However, in my opinion the usage scenarios are very limited, and as such I wouldn't add it to MahApps.Metro. Too many users of MA.M would request more features and adaptability, much more as the control would be able to support.

Alkalinee:
@thoemmi Thank you for your feedback. That a TreeListView is only able to work with items of the same type (or base class) which offers the properties for the columns makes sense, else some columns would always be empty (which would lead to a simple ListView with grouping). But I agree, it's a really special control which won't be used by a lot of developers.

Issue #2652, opened 2016-09-15
[Suggestion] WPF ToolBar - Detect when item is set into ToolBarOverflowPanel


Evangelink:
I wrote a couple of month ago a little helper that can be used on a toolbar item to detect if it is currently displayed or in overflow. The idea was to be able to change the style of the item depending on whether or not it is displayed to reproduce the Windows 10 toolbar (Mail app...).

More info can be found here: http://stackoverflow.com/questions/36328280/wpf-toolbar-detect-when-item-is-set-into-toolbaroverflowpanel

I was wondering if this could be something interesting to incorporate to MahApps? If agreed I will make a PR with a little style example.

Cheers

Evangelink:
Will do the PR as soon as I understand how to commit without having my commit automatically added to the SplitView PR.... *git noob inside*

punker76:
@Evangelink The problem is that you directly work on the develop branch. You should always create your own branch forked from the develop branch to work on a new feature or fixed issue. You can then make a PR from this branch and still working on it without changeing the other PRs.

Evangelink:
@punker76 Ok. Can I undo my PR to recreate it through some branch? Also do you mean 1 branch per fix/feature or just one dev branch and all commits on it?

punker76:
@Evangelink Always One branch from `develop` branch per fix/feature. Maybe you close your current PRs and create new one...

punker76:
I get the same result, if I use the `ToolBar.IsOverflowItem` property, so I’m closing for now this PR.

Please let me know if you still need this extra property.

Thank you! 🙏

PR #2660, opened 2016-09-20
[RFC] Adds a ToolBar helper

Evangelink:
## What changed?

Provides a ToolBar helper. This helper provides a way to tell if the current item is currently being displayed within the overflow panel or within the regular panel. It can be used to create special styles based on the item position.

See issue #2652.


Evangelink:
@punker76 Don't know if it is a bug or just a different mechanism but using MahApps TreeHelper I don't end up with the parent object I am looking for whereas using the given TreeHelper it does work.

avechuche:
Hello! Is this implemented? I can not use it. Thx!

Evangelink:
Hi @punker76 !

I have tried to report my old PR on the up-to-date develop branch. Everything is building ok locally but it seems like something is wrong (references?) on the TeamCity build agent. Not sure whether I missed something (I have updated the shproj by hand cause it doesn't display the xaml pages).

Any help on what I have missed would be appreciated.

punker76:
I get the same result, if I use the `ToolBar.IsOverflowItem` property, so I’m closing for now this PR.

Please let me know if you still need this extra property.

Thank you! 🙏

Evangelink:
Hi @punker76,

I also have the same behavior using `ToolBar.IsOverflowItem`... Either I missed the property or either there was something else broken when I have developed this helper.

Anyway thank you for the time spent on this PR!

Keep up with the good work you are doing!

Issue #2615, opened 2016-08-17
Attached property, call command when reaching end of scroll


Evangelink:
For some applications I wanted to have a UX close to phone apps where I was loading more items when the user was reaching the end of the list (instead of using the old paging system).

I was wondering if it might be of some interest? If so I will do a PR.

PR #2663, opened 2016-09-22
[RFC] Adds an attached property for scrollbar

Evangelink:
## What changed?

This attached property allows to trigger the call to a command when reaching the end of the scrollable area. You can use it for example to create a lazy loading and load more items when the user reach the end of the list.

See #2615


kamilkosek:
Hi,
was looking at it and found a small (maybe copy&paste) typo:
I.e. line 38 :
` public static readonly DependencyProperty EndOfScrollReadchedCommandProperty =` 
but should be 
` public static readonly DependencyProperty EndOfScrollReachedCommandProperty =`

Evangelink:
Hi @kamilkosek! Thanks for this, I will try to fix it as soon as possible!

avechuche:
@Evangelink Hello! Is this implemented? I can not use it. Thx!

Evangelink:
Hi @kamilkosek I made the change (sorry for having been so late).
Hi @avechuche this is implemented in this PR but as you can see this is not merged hence not available on the release MahApps.

@punker76 Were you waiting for something else than a rebase of the PR? Not sure what RFC means.

Cheers,
Amaury

PR #2667, opened 2016-09-23
[RFC] Icon scaling

### LINKED ISSUE #2242: Add padding to window icons
The MetroWindow title bar is not a standard size for icons.  We had an icon that looked sharp with normal windows because it would select the 16x16 icon and add space around the image so it was centered.  The MetroWindow instead scales the image to fit the full square edge to edge, but the title bar isn't big enough to trigger the more detailed 32x32 image that's part of the icon.

As a temporary stop gap I had to create a DrawingImage with a rectangle geometry under the image I wanted.  It allowed me to use the cooler vector images, but this is what my window declaration looked like:

    <controls:MetroWindow>
        <controls:MetroWindow.Icon>
            <DrawingImage>
                <DrawingImage.Drawing>
                    <DrawingGroup>
                    <GeometryDrawing Brush="Transparent">
                        <GeometryDrawing.Geometry>
                            <RectangleGeometry Rect="0,0,72,72"/>
                        </GeometryDrawing.Geometry>
                    </GeometryDrawing>
                    <GeometryDrawing Brush="{StaticResource IconColorBrush}" Geometry="F1 .... Z"/>
                    </DrawingGroup>
                </DrawingImage.Drawing>
            </DrawingImage>
        </controls:MetroWindow.Icon>
    </controls:MetroWindow>

I got the effect I was going for, and if I want I can independently change the foreground and background colors.  However, it would be really nice if I could simply specify an `IconPadding` or an `IconArrange` and specify "Stretch" or "Center" to handle the heavy lifting for me.

thoemmi:
This PR adds the property `IconScalingMode` to `MetroWindow`. It's forwarded to `MultiFrameImage.MultiFrameImageMode` to set the icon scaling. There are two possible values:

* `MultiFrameImageMode.ScaleDownLargerFrame`<br>
  It takes the smallest frame which from the icon which has equal or larger size as the window's icon template. The frame is scaled down it's larger.<br>
  image
* `MultiFrameImageMode.NoScaleSmallerFrame`<br>
  It takes the largest frame from the window which has equal or smaller size than the window's icon template. The frame is rendered centered if it's smaller.<br>
  image

I've added two menu items to the **Window** menu in the demo application.

I created a special icon for reproduction, but didn't want include it in this PR. Just replace *Mahapps.Metro.Tests\mahapps.metro.logo2.ico* with the icon in [this ZIP file](https://github.com/MahApps/MahApps.Metro/files/490773/mahapps.metro.logo2.ico.zip).

**RFC:** I'm not satisfied with the naming of the properties and the enum. If you have a better suggestion I would be happy to rename them.

This fixes #2242.


punker76:
@thoemmi I want to merge this PR with these namings...

For MetroWindow

`IconMultiFrameScalingMode` -> type `IconMultiFrameScalingModes`

For MultiFrameImage

`MultiFrameScalingMode` -> type `IconMultiFrameScalingModes`

```
public enum IconMultiFrameScalingModes {
    ScaleDownLargerFrame,
    NoScaleSmallerFrame,
}
```

thoemmi:
I recommend staying with singular `IconMultiFrameScalingMode`. See [Enumeration Type Naming Guidelines](https://msdn.microsoft.com/en-us/library/4x252001(VS.71).aspx): as long as it's not a bitfield (annotated with `FlagsAttribute`), it should use a singular name, because there's always only one mode active.

The enum is called `MultiFrameImageMode` because it is used by `MultiFrameImage`. That `MetroWindow` is using the control to display the icon is secondary. Therefore I hesitate to prefix the enum's name with `Icon`.

PR #2671, opened 2016-09-26
Proposal - Dont focus flyout on close

ButchersBoy:
### Background

I have multiple windows, and i'm using flyouts to show an alert condition in the application.  When the user dismisses the alert all flyouts are closed via my application code.

### The problem

This line causes the last Window my where my code closes the flyout to receive focus , bringing that window to the front, interrupting the users activity:

```
// focus the Flyout itself to avoid nasty FocusVisual painting (it's visible until the Flyout is closed)		
flyout.Focus();
```

### Remarks

I understand there is a specific comment here regarding this line to perform some visual trickery.  However, it's not logically correct.  I have experimented, removed the line, and I can't see any display issues in _my_ application.

Maybe someone / @punker76 ,  with some background on this line can comment?

Another option would be to make the code a bit more specific; only run the focus if the current Window is active, something like:

```
if (Window.GetWindow(flyout) == Application.Windows.OfType<Window>().FirstOrDefault(w => w.IsActive))
    flyout.Focus();
```

### DIFF
--- a/MahApps.Metro/Controls/Flyout.cs
+++ b/MahApps.Metro/Controls/Flyout.cs
@@ -8,9 +8,7 @@ using System.Windows.Media.Animation;
 using System.Windows.Threading;
 
 namespace MahApps.Metro.Controls
-{
-    using System.Linq;
-
+{    
     /// <summary>
     /// A sliding panel control that is hosted in a MetroWindow via a FlyoutsControl.
     /// <see cref="MetroWindow"/>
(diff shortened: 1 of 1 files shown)


punker76:
@ButchersBoy This hack was introduced with https://github.com/MahApps/MahApps.Metro/pull/1805 by this issue (found by manual testing)

2016-09-26_22h22_18

2016-09-26_22h22_44

But now, I can't also not reproduce this with the current source, so I think we can change this.

ButchersBoy:
Cool! 👍

PR #2688, opened 2016-10-05
Hotfix for DateTimePicker/TimePicker

xxMUROxx:
The reason is that we do not have to use Mode=TwoWay explicit, and DatePicker.SelectedDate has BindsTwoWayByDefault too.

 - [x] Change DateTimePicker.SelectedDate to BindsTwoWayByDefault
 - [x] [Fix] Issue where Binding issues on DateTimePicker/TimePicker are shown on the output window
 - [x] [Fix] Issue where Hour 0 is not shown when using 24h clock
 - [x] Fix issue where clearing text does not immediately set value to `null` but `00:00:00` and `0001.01.01 00.00.00` respectively

### DIFF
--- a/MahApps.Metro/Controls/TimePicker/DateTimePicker.cs
+++ b/MahApps.Metro/Controls/TimePicker/DateTimePicker.cs
@@ -34,6 +34,7 @@
       
         private const string ElementCalendar = "PART_Calendar";
         private Calendar _calendar;
+        private bool _deactivateWriteValueToTextBox;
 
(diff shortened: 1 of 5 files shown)


punker76:
@xxMUROxx I take your changes and put it on the changed source repo -> https://github.com/MahApps/MahApps.Metro/pull/2710
Please take a look to the `DateTimePickerSetCulture`, cause the Text is not set correctly in this case, the time is `00:00:00` in the test... thx

PR #2721, opened 2016-10-28
Remove unnecessary usings

Danghor:
Removes unnecessary using statements to get rid of redundant error messages.


punker76:
@Danghor Can you revert the `Microsoft.Windows.Shell` changes? I want to leave this as is. Thx

PR #2722, opened 2016-10-29
Feature/auto watermark

### LINKED ISSUE #2712: TextBox "Auto" Watermark-text

Not a real issue but a question. 
Would it be possible to automatically set the watermark text to the bound property name ?
Or even better, take automatically the value of the Display-Attribute from the property as the Watermark text.
Lets say I got this Model

``` csharp
public class Device
    {
        [Display(Name = "Geräte Id")]
        public int DeviceId { get; set; }
        [Display(Name = "Geräte Name")]
        public string DeviceName { get; set; }
    }
```

and on the textbox looks e.g. like this

``` xaml
<StackPanel DataContext="{Binding MyDevice}">
                    <TextBox mah:TextBoxHelper.AutoWatermark="True" Text="{Binding DeviceId}"></TextBox>
                </StackPanel>
```

All the best,
Kamil

xxMUROxx:
`AutoWatermark` is able to get the `DisplayAttribute` from the bound property in following cases:
- Binding that is **supported**
  + `"{Binding Path=Property}"`
  + `"{Binding Path=Property.SubProperty}"`
  + `"{Binding Path=CollectionProperty}"`
  + `"{Binding Path=CollectionProperty[0].SubProperty}"`
- Binding that is **not supported**
  + `"{Binding Path=CollectionProperty[0]}"`
 
Tasks
-

 - [x] Searches for WatermarkAttribute on bound property (like stated above)
 - [x] Decision: Is property name adequate
 - [x] Decision: Should error/warning shown if property does not have DisplayAttribute
 - [x] Error must occur if control does not support AutoWatermark

closes #2712


punker76:
@xxMUROxx We can do this for .Net 4 too, look at this here which works...

```csharp
        private static void OnControlWithAutoWatermarkSupportLoaded(object o, RoutedEventArgs routedEventArgs)
        {
            FrameworkElement obj = (FrameworkElement)o;
            obj.Loaded -= OnControlWithAutoWatermarkSupportLoaded;

            DependencyProperty dependencyProperty;

            if (!AutoWatermarkPropertyMapping.TryGetValue(obj.GetType(), out dependencyProperty))
            {
                throw new NotSupportedException($"{nameof(AutoWatermarkProperty)} is not supported for {obj.GetType()}");
            }
            var binding = obj.GetBindingExpression(dependencyProperty);

            if (binding != null)
            {
                var dataItem = binding.DataItem.GetType();
#if NET4_5
                var propertyName = binding.ResolvedSourcePropertyName;
#else
                var propertyName = binding.ParentBinding?.Path?.Path;
#endif
                if (propertyName != null)
                {
                    var property = dataItem.GetProperty(propertyName, BindingFlags.GetProperty | BindingFlags.Public | BindingFlags.Instance);
                    if (property != null)
                    {
#if NET4_5
                        var attribute = property.GetCustomAttribute<WatermarkAttribute>();
#else
                        var attribute = property.GetCustomAttributes(typeof(WatermarkAttribute), false).FirstOrDefault() as WatermarkAttribute;
#endif
                        if (attribute != null)
                        {
                            obj.SetValue(WatermarkProperty, attribute.Caption);
                        }
                    }
                }
            }
        }
#endif
```

But we must make a note to the users, that this feature uses reflection, which can have performance issues if you use this on many controls with watermark in a UI.

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -66,6 +85,61 @@ public static void SetIsSpellCheckContextMenuEnabled(UIElement element, bool val
...
+        {
+            element.SetValue(AutoWatermarkProperty, value);
+        }
+
+        private static void OnAutoWatermarkChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+        {
+            FrameworkElement element = d as FrameworkElement;
+            if (element != null)
+            {
+                element.Loaded += OnControlWithAutoWatermarkSupportLoaded;
```
punker76: Possible memory leak...

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/WatermarkAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public WatermarkAttribute(string caption)
+        {
+            this.Caption = caption;
+        }
+
+        public string Caption { get; set; }
+    }
+#endif
+}
```
punker76: Why is it necessary to introduce a new attribute? Is the Display attribute too bad?

xxMUROxx: No, of course, it is not.

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
xxMUROxx: Because of the event subscription? Shouldn't it be enough to unsubscribe directly after the event was raised the first time?

xxMUROxx:
@punker76 XAML-Style change was introduces unintentionally. I will revert them

xxMUROxx:
@punker76 reverted xaml, added .net4 support and hopefully removed memory leak

punker76:
@xxMUROxx It seems this feature is done. So can I merge it?

xxMUROxx:
@punker76 yes

kamilkosek:
Thanks guys

Issue #2729, opened 2016-11-01
[Enhancement] Don't remove window border if MaxWidth/MaxHeight < Resolution on maximize


Deadpikle:
## What steps will reproduce this issue?

1) Set MaxHeight/MinHeight to < resolution of monitor
2) "Maximize" application by dragging to top of screen or by clicking the maximize button
3) Window border is gone!

If MaxWidth/MaxHeight is < the current resolution of the monitor, the application can still be "maximized" by dragging the window to the top of the monitor. This happens in stock WPF apps and is OK. However, I think the window border shouldn't be removed in this case. It isn't removed in stock WPF apps.

Comparison screenshot:
![comparison](https://cloud.githubusercontent.com/assets/5092399/19891858/64fa128e-a018-11e6-867e-a773e6114869.png)

The top window is a stock WPF app; the lower window is a MetroWindow. Note that both windows were "maximized" by dragging the title bar to the top of the screen.

### Code

Stock WPF:
```
<Window x:Class="DataDrivenMenusSample.Window1"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    Title="Window1" MaxHeight="300" MaxWidth="300">
    <Grid>
    </Grid>
</Window>
```

MetroWindow:
```
<Controls:MetroWindow x:Class="MaximizeWindowTest.MainWindow"
                      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                      xmlns:Controls="http://metro.mahapps.com/winfx/xaml/controls"
                      Title="MainWindow"
                      Height="400"
                      Width="400"
        BorderBrush="{DynamicResource AccentColorBrush}"
        BorderThickness="1"
        TitleCaps="False"
        WindowButtonCommandsOverlayBehavior="HiddenTitleBar"
        UseNoneWindowStyle="False"
        MaxHeight="400" 
        MaxWidth="400"
        WindowStartupLocation="CenterScreen"
        WindowTransitionsEnabled="False"
        Opacity="1"
        ShowDialogsOverTitleBar="False">
    <Grid>
    </Grid>
</Controls:MetroWindow>
```

### Expected outcome

[...]

PR #2731, opened 2016-11-01
Keep BorderThickness on maximize if necessary

Deadpikle:
## What changed?

Implements enhancement #2729 

Adds `KeepBorderOnMaximize` property to `MetroWindow`. This property defaults to `False` so that old behavior is maintained. If `True`, the right/bottom border is kept on maximize if the maximized window size will be less than the current monitor resolution.

This screenshot shows what the demo window looks like when `MaxHeight` and `MaxWidth` are both 400 and `KeepBorderOnMaximize` is `True`.
topleft

Note that I had to reorder the logic just slightly in `HandleMaximize()` so that I could use the `monitorInfo` variable earlier.


Deadpikle:
Closing due to fix being in ControlzEx now.

PR #2758, opened 2016-12-06
Added more TextBox ContextMenu configuration

Deadpikle:
## What changed?

- Allow for disabling cut/copy/paste in a TextBox
- Allow for custom context menu items in a TextBox
- Added IsCutCopyPasteInContextMenu property (defaults to true)
- Added ExtraContextMenuItems property (defaults to null)

I'm not sure that properties were the best way to go about doing this, but I thought I'd do a pull request all the same. I'm open to better ways.

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
+++ b/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
@@ -15,6 +15,9 @@ using JetBrains.Annotations;
 
 namespace MahApps.Metro.Controls
 {
+    using System.Collections.Generic;
+    using System.ComponentModel;
+
     /// <summary>
     /// A helper class that provides various attached properties for the TextBox control.
     /// </summary>
@@ -53,6 +56,8 @@ namespace MahApps.Metro.Controls
         public static readonly DependencyProperty HasTextProperty = DependencyProperty.RegisterAttached("HasText", typeof (bool), typeof (TextBoxHelper), new FrameworkPropertyMetadata(false, FrameworkPropertyMetadataOptions.AffectsMeasure | FrameworkPropertyMetadataOptions.AffectsArrange | FrameworkPropertyMetadataOptions.AffectsRender));
 
         public static readonly DependencyProperty IsSpellCheckContextMenuEnabledProperty = DependencyProperty.RegisterAttached("IsSpellCheckContextMenuEnabled", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(false, UseSpellCheckContextMenuChanged));
+        public static readonly DependencyProperty IsCutCopyPasteInContextMenuProperty = DependencyProperty.RegisterAttached("IsCutCopyPasteInContextMenu", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(true, IsCutCopyPasteInContextMenuChanged));
+        public static readonly DependencyProperty ExtraContextMenuItemsProperty = DependencyProperty.RegisterAttached("ExtraContextMenuItems", typeof(List<MenuItem>), typeof(TextBoxHelper), new FrameworkPropertyMetadata(null, ExtraContextMenuItemsChanged));
 
         /// <summary>
         /// This property can be used to retrieve the watermark using the <see cref="DisplayAttribute"/> of bound property.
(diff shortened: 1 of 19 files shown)


src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -247,24 +252,60 @@ private static void UseSpellCheckContextMenuChanged(DependencyObject d, Dependen
...
+                throw new InvalidOperationException("The property 'ExtraContextMenuItemsChanged' may only be set on TextBoxBase elements.");
+            }
+            else
+            {
+                tb.SetValue(ExtraContextMenuItemsProperty, e.NewValue);
+                tb.ContextMenu = GetDefaultTextBoxBaseContextMenu(tb);
+                AddExtraItemsToContextMenu(tb, tb.ContextMenu.Items.Count > 0);
+                if (tb.ContextMenu.Items.Count == 0)
+                {
+                    AddNoItemsAvailToContextMenu(tb);
```
punker76: Why should we add a menu with no function behind?

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -53,6 +56,8 @@ public class TextBoxHelper
         public static readonly DependencyProperty HasTextProperty = DependencyProperty.RegisterAttached("HasText", typeof (bool), typeof (TextBoxHelper), new FrameworkPropertyMetadata(false, FrameworkPropertyMetadataOptions.AffectsMeasure | FrameworkPropertyMetadataOptions.AffectsArrange | FrameworkPropertyMetadataOptions.AffectsRender));
 
         public static readonly DependencyProperty IsSpellCheckContextMenuEnabledProperty = DependencyProperty.RegisterAttached("IsSpellCheckContextMenuEnabled", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(false, UseSpellCheckContextMenuChanged));
+        public static readonly DependencyProperty IsCutCopyPasteInContextMenuProperty = DependencyProperty.RegisterAttached("IsCutCopyPasteInContextMenu", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(true, IsCutCopyPasteInContextMenuChanged));
```
punker76: Isn't it better to name it think `UseDefaultContextMenu` or `AllowDefaultContextMenu`...
/cc @thoemmi

Deadpikle: I like `UseDefaultContextMenu` much better than what I chose! I'll wait a day or two before changing it to wait to hear from @thoemmi .

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
Deadpikle: (let me know if you want me to rebase)

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -301,23 +336,48 @@ private static void TextBoxBaseContextMenuOpening(object sender, ContextMenuEven
...
+            tbBase.ContextMenu.Visibility = tbBase.ContextMenu.Items.Count > 0 ? Visibility.Visible : Visibility.Hidden;
+        }
+
+        private static void AddExtraItemsToContextMenu(TextBoxBase tbBase, bool shouldAddSeparator = false)
+        {
+            List<MenuItem> extraMenuItems = (List<MenuItem>)tbBase.GetValue(ExtraContextMenuItemsProperty);
+            if (extraMenuItems != null && extraMenuItems.Count > 0)
+            {
+                // Add separator if necessary, then extra items to menu
+                if (shouldAddSeparator)
```
punker76: I think we should not add a default separator, cause users can't avoid this. If the user want it then he should do this itself.

Deadpikle: The `List<MenuItem>` is now a `List<FrameworkElement>`.

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -247,24 +252,54 @@ private static void UseSpellCheckContextMenuChanged(DependencyObject d, Dependen
...
 
+        private static void IsCutCopyPasteInContextMenuChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+        {
+            var tb = d as TextBoxBase;
+            if (null == tb)
+            {
+                throw new InvalidOperationException("The property 'IsCutCopyPasteInContextMenuChanged' may only be set on TextBoxBase elements.");
+            }
+
+            tb.SetValue(IsCutCopyPasteInContextMenuProperty, e.NewValue);
```
thoemmi: Is this really necessary? I'd think this method is called **because** the `IsCutCopyPasteInContextMenu` was changed, so setting it again is not needed. But I may be wrong.

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -247,24 +252,54 @@ private static void UseSpellCheckContextMenuChanged(DependencyObject d, Dependen
...
+        {
+            var tb = d as TextBoxBase;
+            if (null == tb)
+            {
+                throw new InvalidOperationException("The property 'ExtraContextMenuItemsChanged' may only be set on TextBoxBase elements.");
+            }
+            else
+            {
+                tb.SetValue(ExtraContextMenuItemsProperty, e.NewValue);
+                tb.ContextMenu = GetDefaultTextBoxBaseContextMenu(tb);
```
thoemmi: Updating the context menu in each and every `XxxChanged` method seems wrong to me. I'd think that creating the actual context menu inside `TextBoxBaseContextMenuOpening` should be sufficient. @punker76 ?

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
thoemmi: I agree with @punker76 that the naming can be improved. However, I'd would negate it (generally I think boolean properties should be `false` by default and you have to switch 'em on explicitly) and name it `ExcludeDefaultContextMenuItems`.

punker76: 👍

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
punker76: @thoemmi You're not wrong, this is not necessary. /cc @Deadpikle

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
punker76: @thoemmi Yes the opening event should be sufficient /cc @Deadpikle So the ExtraContextMenuItemsChanged is not necessary.

thoemmi:
Ok, I see the point. There's already a `TextBoxBaseContextMenuOpening`, but
it's used only in `UseSpellCheckContextMenuChanged`. Adding the
registration to your new `XxxChanged` methods wouldn't help, because if you
would call `UseSpellCheckContextMenu="false"`, the event would get
unsubscribed.

As far as I can see now is that having all the `XxxChanged` are required.
However, move the creation of the `ContextMenu` to a single method, so it
can be called from several places.

Deadpikle:
@punker76 @thoemmi Sorry for the wait. Refactored the context menu creation code to a single method. I added a private property to help keep track of the `ContextMenuOpening` event so it isn't added more than one time.
Let me know if there are any further changes you'd like to see.

punker76:
@Deadpikle Hey, thx for this changes and your time to do this. Sorry for my lateness. I review this and had some thoughts about this and the new attached properties. And I come to the conclusion that we don't need this new properties with some other changes...
Here are my changes to solve this without adding new attached properties https://github.com/MahApps/MahApps.Metro/pull/2772.
I hope this is ok for you 🙈 
/cc @thoemmi

Deadpikle:
@punker76  It's cool; I'm glad there is a solution that will work for everyone and is a lot cleaner than what I did. :)

PR #2763, opened 2016-12-13
[Breaking change] DateTimePicker/TimePicker now using DateTime?

xxMUROxx:
@punker76 @michiwieland


xxMUROxx:
For now I have changed the properties to `SelectedDateTime`. Is this OK, or should the name be different?

PR #2769, opened 2016-12-17
Expander fix closing animation

Alkalinee:
Fix the closing animation of the expander, it was not executed because the expander was directly collapsed. The closing animation duration was also made smaller because 250 ms was too long (changed to 100 ms, better in my feeling).


punker76:
@Alkalinee Thx! I revert the duration for the collapse animation to 250. But you can simply change it now by yourself in your app like this

```
<!--  quick change of the collapse storyboard duration  -->
<Storyboard x:Key="MetroExpanderCollapseStoryboard">
    <DoubleAnimation Duration="0:0:0.1"
                        Storyboard.TargetName="ExpandSite"
                        Storyboard.TargetProperty="(UIElement.Opacity)"
                        To="0" />
    <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ExpandSite" Storyboard.TargetProperty="(UIElement.Visibility)">
        <DiscreteObjectKeyFrame KeyTime="0:0:0.1" Value="{x:Static Visibility.Collapsed}" />
    </ObjectAnimationUsingKeyFrames>
</Storyboard>
```

punker76:
@Alkalinee I revert the dynamic resource search for the Storyboard, cause if you don't put the MahApps resources to the App.xaml you will get a XamlParseException. So for now it's not so easy to change the duration for this Storyboard. I'll investigate for next releases what we can do...

PR #2774, opened 2016-12-22
Fix DataGridRowHeader HorizontalAlignment

chuuddo:
## What changed?
Some time ago i have made pull request #1819 that adds alignments to DataGridRowHeader, but in 6df54bab96e82be811edd414826636ca9084ba94 this feature was broken.

Before:
before
After:
after


punker76:
@chuuddo 👍

Issue #2585, opened 2016-07-16
HotKeyBox gives wrong names to multimedia buttons


bigworld12:
## What steps will reproduce this issue?

when assigning a multimedia key to the HotKeyBox it will display it with wrong name e.g.
Key.MediaPlayPause > shows like this : http://puu.sh/q3BTW/da94bd8c0d.png
![G Instead of MediaPlayPause](http://puu.sh/q3BTW/da94bd8c0d.png)
### Expected outcome

it should view MediaPlayPause instead of G
## 
### Environment
- MahApps.Metro **v1.3.0-ALPHA176**
- Windows **10**
- Visual Studio **2015**
- .NET Framework **4.5.2**

thoemmi:
I can reproduce the issue. However, I don't know how to fix it at the moment. [`GetKeyNameText`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms646300(v=vs.85).aspx) just returns `G` for `MediaPlayPause`. I'll have to investigate further...

bigworld12:
Why not use Key.ToString() instead ?

thoemmi:
Unfortunately, the Windows API can't help. I asked at [StackOverflow](http://stackoverflow.com/q/38584015/4747), and there's no standard way to get the names of special keys such as the media buttons. 😫

bigworld12:
then maybe create a simple dictionary of Key,string for special keys ,and then we can handle localization later ?

PR #2814, opened 2017-01-24
Correct display of media buttons in HotKeyBox

thoemmi:
This PR fixes the wrong display of special media buttons in `HotKeyBox` as described in #2585.

Unfortunately, Windows does not localize media buttons, therefore I've implemented a fallback using `ToString()`.

Additionally, upper case isn't hard-coded anymore. The user has now the option to use `ControlsHelper.ContentCharacterCasing`. (Upper case is still the default)

I haven't edited any release notes in the PR as I don't know in which release will include this PR.


thoemmi:
I don't think that the failed build is caused by my changes in any way 🤞

PR #2829, opened 2017-02-01
logo as xaml and svg

ButchersBoy:
Logo in vector form; XAML and SVG.


punker76:
@ButchersBoy thx mate 👍 👏

PR #2837, opened 2017-02-03
[Fix] Issue when the time is cleared if a date is selected

xxMUROxx:
## What changed?

Before | After
------------ | -------------
**Time is set to 00:00:00** | **Time is now kept**
datetimepicker-time-before | datetimepicker-time-after
**If a date is selected then selecting the time closes the control** | **The control does now stays open**
datetimepicker | datetimepicker-fixfocus


xxMUROxx:
@punker76 this PR can be merged, if desired

PR #2851, opened 2017-02-12
Code clone reduction

manne:
## What changed?

The code base has some clones.
This PR removes the clone group of the external dialog functionality.

### Before 
devenv_2017-02-12_16-13-11


manne:
less code
image

punker76:
@manne I am with you, code clones are not good. But this PR is really too big to check if all features still work. You broke the tests with your last changes. So it's better to make smaller PRs and check if the tests still running and no feature is gone and still work. thx

PR #2854, opened 2017-02-13
window's default behavior : double click on title bar will restore Minimized window

hausenism:
If ShowInTaskbar = false, when double click on the title bar of a Minimized window, the window will be Restored instead of Maximized.


punker76:
@hausenism Thx! 👍

PR #2858, opened 2017-02-16
Fix WindowCommands tab stop bug

neilt6:
## What changed?

Prevents `WindowCommandsItem` from erroneously responding to the tab key.

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro/Themes/WindowCommands.xaml
+++ b/src/MahApps.Metro/MahApps.Metro/Themes/WindowCommands.xaml
@@ -115,6 +115,7 @@
         <Setter Property="ItemContainerStyle">
             <Setter.Value>
                 <Style TargetType="{x:Type Controls:WindowCommandsItem}">
+                    <Setter Property="IsTabStop" Value="False"/>
                     <Setter Property="Template">
                         <Setter.Value>
                             <ControlTemplate TargetType="{x:Type Controls:WindowCommandsItem}">


punker76:
@thoemmi What do you think?

neilt6:
`WindowCommandItem` is acting as a tab stop, but doesn't pass keyboard input to the `Button` it contains, making it useless behavior that can't be disabled from user code. Even if it did work, I find it counter-intuitive since the usual window commands _(minimize, maximize, close, or even back in the case of UWP)_ aren't normally tab stops.

thoemmi:
The general window commands Minimize, Maximize, and Close don't get the focus with MahApps.Metro too. Those are generic Windows commands and not focusable in any Windows application (as far as I'm aware of.)

However, the additional Windows commands provided by MahApps.Metro are specific to the application, and IMHO should be accessible via keyboard.

If the buttons don't work, I'd think that this is the issue that must be fixed, and not by preventing them from getting the focus at all.

neilt6:
I assumed this was a bug since `WindowCommands.IsTabStop` is set to `false` in the default style, and the buttons aren't actually interactive. In addition to fixing the tab stop behavior on the buttons, we should also make `WindowCommands.IsTabStop` enable/disable all of the tab stops like you would expect.

punker76:
@neilt6 The inner buttons of the WindowCommandsItem are now focusable and interact with user key input.
/cc @thoemmi can you check the result of my last changes? thx

punker76:
@neilt6 -> 1.5.0-alpha002

thoemmi:
@punker76 Looks good to me

punker76:
@thoemmi thx for watching 👍

thoemmi:
Wow, it took you 30 seconds to reply to my comment 😉

PR #2868, opened 2017-02-24
Don't overwrite cancellation for close

batzen:
## What changed?

If someone listening for `Closing` set `e.Cancel = true` we should not overwrite/overrule that request.


punker76:
@batzen thx Bastian

Issue #2889, opened 2017-03-13
[Feature Request] Watermark Wrapping


amkuchta:
## What steps will reproduce this issue?

_Similar to #2884, any watermark applied to a `TextBox` or `ComboBox` control will not wrap when the control becomes too short to display the information_

### Expected outcome

_The watermark should follow the same behavior as the `TextBox` - that is, if the `TextBox` has `TextWrapping="Wrap"`, the watermark should have the same setting_

Issue #2884, opened 2017-03-09
[Feature Request] Watermark Trimming


amkuchta:
## What steps will reproduce this issue?

_Currently, any watermark applied to a `TextBox` or `ComboBox` control will not trim when the control becomes too short to display the information_

### Expected outcome

_Text should trim at the character/word level, followed by a set of ellipses_

amkuchta:
Personally, I think that the watermark should follow the same setting as the `TextBox` instead of having to set a separate property (if possible) @xxMUROxx, thoughts?

xxMUROxx:
Yes, that would be a good option@punker76Am 13.03.2017 14:43 schrieb Alex Kuchta <notifications@github.com>:Personally, I think that the watermark should follow the same setting as the TextBox instead of having to set a separate property (if possible) @xxMUROxx, thoughts?

amkuchta:
@xxMUROxx once I get #2892 done and approved / rejected, I will look at tackling this and #2889 (I added commits to the above PR that shouldn't have been there 💩 , so now I want to get it taken care of before I tackle anything else)

xxMUROxx:
@amkuchta You can remove the commits that are wrong on the branch representing the PR. Then you can `git push -f` to update the PR. Since no merge is done yet, there is no problem if you rewrite history.

amkuchta:
@punker76 do you want me to roll this and #2889 into #2898 (I would rename the PR to "Watermark Enhancements", but this seems to fit there), or are you actively working it now?

@xxMUROxx thanks for the advice - I separated the PR, so now I can work this without any guilt 😄

amkuchta:
@xxMUROxx I mixed the `TextBox` and `TextBlock` controls up... `TextBox` does not have a `TextTrimming` property.  I am integrating a new DependencyProperty to handle this.

amkuchta:
@xxMUROxx @punker76 So, if I have both `TextTrimming` and `TextWrapping` set for the watermark, `TextWrapping` overrides `TextTrimming`.  Should we keep it this way, or should I make `TextTrimming` the dominant behavior?

PR #2898, opened 2017-03-17
[RFC] [WIP] Watermark Enhancements

amkuchta:
## What changed?

I love the functionality of the floating watermark, but don't like that the border stretches, as do any buttons contained within the control. My solution is to implement a watermark that floats outside of the control.  Improvements include:

- Creation of a DependencyProperty, `FloatingWatermarkLocation`, which can be used on the following controls:
  - TextBox
  - ComboBox
  - DatePicker
  - DateTimePicker
  - PasswordBox
  - NumericUpDown
- Creation of an enum named FloatingWatermarkLocations to outline values acceptable for the newly created dependency property: Interior and Exterior
- Updates to the view to demonstrate the new feature (TextBox, DatePicker, DateTimePicker, ComboBox)
- Implements `Controls:TextBoxHelper.FloatingWatermarkOverrideText`, which allows users to set a custom text for floating watermarks if they feel that the current watermark is too long

## TODO (Help Needed)
Status on all controls:

- [x] TextBox - Fully implemented
- [x] PasswordBox - Fully implemented
- [x] DatePicker - Fully implemented
- [x] DateTimePicker - Fully implemented
- [ ] ComboBox - Partially implemented: the `FloatingWatermarkLocation` works well on non-editable ComboBox controls... but if editing is enabled, forget about it (it works on initial load, but I cannot get it to work after the text has been changed). I could use some help wiring this together
- [ ] NumericUpDown - No implementation: The Dependency Property can be used for the control, but for some reason, the `TemplateBinding` isn't working...

@thoemmi , I figured that you may be the best person to take a look, since you did the initial creation of the UseFloatingWatermark feature. Thoughts? (Sorry for the duplicate tag - I moved this to its own PR, where it belongs)

## Closed Issues
- [x] #2884
- [x] #2889

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
+++ b/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
@@ -32,6 +32,7 @@ namespace MahApps.Metro.Controls
     }
 
     public enum FloatingWatermarkLocations { Interior, Exterior };
+    public enum WatermarkTrimmingValues { CharacterEllipsis, None, WordEllipsis };
 
     /// <summary>
     /// A helper class that provides various attached properties for the TextBox control.
@@ -44,7 +45,8 @@ namespace MahApps.Metro.Controls
         public static readonly DependencyProperty IsMonitoringProperty = DependencyProperty.RegisterAttached("IsMonitoring", typeof(bool), typeof(TextBoxHelper), new UIPropertyMetadata(false, OnIsMonitoringChanged));
         public static readonly DependencyProperty WatermarkProperty = DependencyProperty.RegisterAttached("Watermark", typeof(string), typeof(TextBoxHelper), new UIPropertyMetadata(string.Empty));
         public static readonly DependencyProperty UseFloatingWatermarkProperty = DependencyProperty.RegisterAttached("UseFloatingWatermark", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(false, ButtonCommandOrClearTextChanged));
-        public static readonly DependencyProperty FloatingWatermarkLocation = DependencyProperty.RegisterAttached("FloatingWatermarkLocation", typeof(FloatingWatermarkLocations), typeof(TextBoxHelper), new FrameworkPropertyMetadata(FloatingWatermarkLocations.Interior, ButtonCommandOrClearTextChanged));
+        public static readonly DependencyProperty FloatingWatermarkLocationProperty = DependencyProperty.RegisterAttached("FloatingWatermarkLocation", typeof(FloatingWatermarkLocations), typeof(TextBoxHelper), new FrameworkPropertyMetadata(FloatingWatermarkLocations.Interior, ButtonCommandOrClearTextChanged));
+        public static readonly DependencyProperty WatermarkTrimmingProperty = DependencyProperty.RegisterAttached("WatermarkTrimming", typeof(WatermarkTrimmingValues), typeof(TextBoxHelper), new FrameworkPropertyMetadata(WatermarkTrimmingValues.CharacterEllipsis));
         public static readonly DependencyProperty TextLengthProperty = DependencyProperty.RegisterAttached("TextLength", typeof(int), typeof(TextBoxHelper), new UIPropertyMetadata(0));
         public static readonly DependencyProperty ClearTextButtonProperty = DependencyProperty.RegisterAttached("ClearTextButton", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(false, ButtonCommandOrClearTextChanged));
         public static readonly DependencyProperty TextButtonProperty = DependencyProperty.RegisterAttached("TextButton", typeof(bool), typeof(TextBoxHelper), new FrameworkPropertyMetadata(false, ButtonCommandOrClearTextChanged));
@@ -439,12 +441,29 @@ namespace MahApps.Metro.Controls
         [AttachedPropertyBrowsableForType(typeof(TimePickerBase))]
         public static string GetFloatingWatermarkLocation(DependencyObject obj)
         {
-            return obj.GetValue(FloatingWatermarkLocation).ToString();
+            return obj.GetValue(FloatingWatermarkLocationProperty).ToString();
         }
 
         public static void SetFloatingWatermarkLocation(DependencyObject obj, FloatingWatermarkLocations value)
         {
-            obj.SetValue(FloatingWatermarkLocation, value);
+            obj.SetValue(FloatingWatermarkLocationProperty, value);
+        }
+
+        [Category(AppName.MahApps)]
+        [AttachedPropertyBrowsableForType(typeof(TextBoxBase))]
+        [AttachedPropertyBrowsableForType(typeof(PasswordBox))]
+        [AttachedPropertyBrowsableForType(typeof(ComboBox))]
+        [AttachedPropertyBrowsableForType(typeof(DatePicker))]
+        [AttachedPropertyBrowsableForType(typeof(TimePickerBase))]
+        [AttachedPropertyBrowsableForType(typeof(NumericUpDown))]
+        public static string GetWatermarkTrimming(DependencyObject obj)
+        {
+            return obj.GetValue(WatermarkTrimmingProperty).ToString();
+        }
+
+        public static void SetWatermarkTrimming(DependencyObject obj, WatermarkTrimmingValues value)
+        {
+            obj.SetValue(WatermarkTrimmingProperty, value);
         }
 
         /// <summary>
--- a/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
+++ b/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
@@ -31,6 +31,8 @@ namespace MahApps.Metro.Controls
     {
     }
 
+    public enum FloatingWatermarkLocations { Interior, Exterior };
+
     /// <summary>
     /// A helper class that provides various attached properties for the TextBox control.
     /// </summary>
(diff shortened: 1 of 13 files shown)


amkuchta:
@punker can I ask why the "Breaking" label?  I took care to ensure that "Interior" is the default setting - users don't have to declare it, and their code will work fine following the update, as demonstrated in the MetroDemo in my branch (unless you think something in `ComboBox` or `NumericUpDown` will break existing functionality)

thoemmi:
Reviewing this PR is hard as it also reformats many XAML files. @amkuchta please consider using [XAML Styler](https://github.com/Xavalon/XamlStyler/), a VS extension for formatting XAMl files. @punker76 [added a settings file](https://github.com/MahApps/MahApps.Metro/commit/f0517b00561ac4a7bd6e625b6308c624b6f93419) for XAML Styler.

amkuchta:
@thoemmi no problem, I'll pull the settings file. I already use XAML Style for my other projects, but my formatting for them is different. I'll ingest the settings for MA.M and add another commit later today (US time)

amkuchta:
@thoemmi @punker76  formatting is updated per the referenced XAML Styler settings file - should be easier to follow along with now.  I also began implementing  `Controls:TextBoxHelper.WatermarkTrimming` (#2884), but for some reason `TemplateBinding Controls:TextBoxHelper.WatermarkTrimming` is not overriding the default style on `PART_Message`.  It is the only DependencyProperty that is not working.

This also implements #2889 by applying `TextWrapping={TemplateBinding TextWrapping}` on `PART_Message` for `TextBox` controls - I still need to add this to the other controls.

punker76:
I must really say make changes always without any code formatting changes together, cause this makes it always hard to read and review the chnages...

amkuchta:
@punker76 noted. Might not be a bad idea to add text about the XAML Style settings file in the PR template - it would probably help to make sure that everyone is on the same page when they submit

punker76:
@amkuchta will do this, sorry

amkuchta:
@punker76 no need to apologize - just trying to help keep guys like me from creating headaches for guys like you

amkuchta:
`Controls:TextBoxHelper.WatermarkTrimming` now works - I realized that creating my own `enum` was the issue, I need to use the provided `System.Windows.TextTrimming`.

I'll incorporate this throughout (including on floating watermarks) tonight / tomorrow

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -31,6 +31,8 @@ public class SpellCheckSeparator : Separator, ISpellCheckMenuItem
     {
     }
 
+    public enum FloatingWatermarkLocations { Interior, Exterior };
```
xxMUROxx: Could you consider renaming this enum to singular. Because plural enum name should be used for enums with a [FlagAttribute]

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -424,7 +430,41 @@ public static void SetUseFloatingWatermark(DependencyObject obj, bool value)
...
-        
+
+        [Category(AppName.MahApps)]
+        [AttachedPropertyBrowsableForType(typeof(TextBoxBase))]
+        [AttachedPropertyBrowsableForType(typeof(PasswordBox))]
+        [AttachedPropertyBrowsableForType(typeof(ComboBox))]
+        [AttachedPropertyBrowsableForType(typeof(NumericUpDown))]
+        [AttachedPropertyBrowsableForType(typeof(DatePicker))]
+        [AttachedPropertyBrowsableForType(typeof(TimePickerBase))]
+        public static string GetFloatingWatermarkLocation(DependencyObject obj)
```
xxMUROxx: Why returning a string when FloatingWatermarkLocation is of type `FloatingWatermarkLocations`?

src/MahApps.Metro/MahApps.Metro.Shared/Controls/Helper/TextBoxHelper.cs
```diff
@@ -424,7 +430,41 @@ public static void SetUseFloatingWatermark(DependencyObject obj, bool value)
...
+        }
+
+        [Category(AppName.MahApps)]
+        [AttachedPropertyBrowsableForType(typeof(TextBoxBase))]
+        [AttachedPropertyBrowsableForType(typeof(PasswordBox))]
+        [AttachedPropertyBrowsableForType(typeof(ComboBox))]
+        [AttachedPropertyBrowsableForType(typeof(DatePicker))]
+        [AttachedPropertyBrowsableForType(typeof(TimePickerBase))]
+        [AttachedPropertyBrowsableForType(typeof(NumericUpDown))]
+        public static string GetWatermarkTrimming(DependencyObject obj)
```
xxMUROxx: Why returning a string when WatermarkTrimming is of type `TextTrimming`?

amkuchta: @xxMUROxx  I will rename the `enum`, and the two functions return `string` because `FloatingWatermarkLocation` started out as `typeof string` and I forgot to change it :laughing: I'll change this too - good catch!

xxMUROxx: 👍

amkuchta:
@xxMUROxx updated to address your concerns with the latest commit

amkuchta:
Thought of another item to implement that is Watermark based: an ability to override longer watermarks with something shorter for display when FloatingWatermarks are in use (I need to change it to a `MultiDataTrigger` that also takes into account `UseFloatingWatermarks`, but this works for a demo purpose).  Notes:

- If `FloatingWatermarkOverrideText` is set, Floating watermark becomes the set text
- If `FloatingWatermarkOverrideText` is not set, the value provided for `Watermark` is used for the floating values (same as before)
- This is not a breaking change

Example GIF:

floatingoverride

amkuchta:
As a side note, if you all (looking at you @thoemmi and @punker76) until after I have finished implementing things & stuff, I can let you know.  I should only have one commit left, though (excluding any changes you guys come across)

amkuchta:
@punker76 @thoemmi I'm done!  I updated the main text for the PR to reflect what is done and what still needs to happen - short version is that editable `ComboBox` controls and `NumericupDown` still are not working with the `FloatingWatermarkLocation`.  I have updated the views to show some implementation, and once these are approved, I will work them into my proposed changes for the MetroDemo.

Let me know if you guys want me to change anything, and thanks in advance for the help with the `ComboBox` and `NumericUpDown` issues!

xxMUROxx:
You have two options:- merge develop into this branch (easiest) - rebase the branch on develop (ou have to force push) I personally prefer the first as you can see on open PR created by meAm 20.04.2017 17:03 schrieb Alex Kuchta <notifications@github.com>:@punker76 @thoemmi I see now that there are some conflicts with this PR and the develop branch now that v1.5.0 is out, and I know that you guys were having some issues with completing the reviews because of the code formatting.  Would it make things easier on you guys if I closed this PR, moved the changes to a newer copy of develop, and resubmitted?  I don't particularly want to have to redo the work, but if it helps you guys out (which, in turn, helps me by getting the changes approved faster), I am willing to do it.
Thoughts?

amkuchta:
@xxMUROxx, the issue is that the history itself is messed up - changes that I made were overwritten by XAML formatting updates, which makes analyzing the diff hard for @punker76 and @thoemmi. Creating a new pull request (which has the correct formatting applied from the start) will make analyzing the diff easier, which speeds up the review process.

amkuchta:
@thoemmi `FloatingWatermarkOverrideText` is a separate feature from `FloatingWatermarkLocation` - the two can be used independently.  Consider the following example:

I have a watermark that reads _"Email Address (you@example.com)"_.  I want the example in the watermark, I only want my floating watermark to read "Email Address" (without the example, because it looks ugly).  To achieve this, I set `FloatingWatermarkOverrideText="Email Address"`.  Now, I have the example in the TextBox, and the "pretty" version floating, when the example is no longer required because the TextBox is populated.

I could then go a step further by deciding to have the watermark float exterior to the box by setting `FloatingWatermarkLocation="Exterior"`, though this is not necessary if I am happy with having the floating watermark inside of the TextBox control.

Hope that clarifies!

thoemmi:
Ok, I see your point. However, **personally** I don't like watermark's appearing and disappearing, regardless if inside or outside the textbox: it changes the layout, and the user might be irritated by the moving controls. Like I said in #1825, what you're talking about is a header and not a watermark.

amkuchta:
@thoemmi the watermark shouldn't be "appearing or disappearing" - the transitions are the same, it's really just a matter of where the watermark is being shown and what it says.

To address the "header" topic, wouldn't the current floating watermark really be just a header that is inside of the border of the TextBox?  And as far as the moving control goes, the `FloatingWatermark` already shifts layout when it expands the TextBox - this is no different, just that the TextBox itself is not expanding in size, simply shifting downward to make room for the new watermark / header.

In any case, both points are semi-moot; the changes I have made use your original settings as the standard, and users must explicitly declare my options in order for them to take effect.  If the users like yours better, they do not have to adjust their implementation at all, and current users will not see any change in how the controls function following the incorporation of this feature.

thoemmi:
Indeed, I'm not a fan of the already existing `FloatingWatermark`,  so my personal opinion is not only about your PR 😉 

I'm curious how many users of MahApps.Metro are really using `FloatingWatermark`. I'd guess it's only a very small fraction. The `MetroTextBox` style is getting out of hand, and the maintainability will be suffer with every additional "exotic" feature. I'm sure your request is reasonable in your scenario, but I'm afraid that someday we'll have a "feature explosion". IMHO if a requirement is too special, you better use a custom style in your own application instead of passing it on to Mahapps.Metro (or any other generic library), which must serve lots of different use cases.

Anyway, that's just my $ 0.02. At the end of the day @punker76 has the final say.

amkuchta:
@thoemmi it's good to know that I shouldn't take it personally!  In all honesty, I understand the "feature creep" that you are concerned about - unmanageable code is a pain.  I wanted to implement this at the MahApps level for two reasons:

1. Others may find it useful or may like my style just a bit better, which could up the "adoption" of the floating watermark
2. (And more importantly) it was easier for me to make the changes at the library level as opposed to a custom implementation, as all of the code is in place for transitions, etc.

Like most devs, I'm lazy :rofl:

amkuchta:
Closing to create (another) new PR that is cleaner and has the correct formatting... It'll be a bit, largely because git sucks.

PR #2899, opened 2017-03-17
Button.xaml - Change ControlsHelper.CornerRadius to ButtonHelper.CornerRadius

waydec:
I realize the plan is to deprecate ButtonHelper but the style causes an error in the xaml designer.

Designer now displays properly rendered control


punker76:
@waydec what is the designer error?

PR #2943, opened 2017-04-25
DependencyProperty DisplayFormat for DateTimePicker and TimePicker

mosberger:
Customizable format of the displayed Date / DateTime.
The default value is still the format of the SpecificCultureInfo.


mosberger:
Well, I think this change has already been made with #2930.

punker76:
@xxMUROxx Can we maybe merge this together with #2930 ?

mosberger:
@punker76 I think the solution of @xxMUROxx  is more complete than mine. You can delete my pull request. 
Sorry for not checking the open pull requests.

xxMUROxx:
@punker76 @mosberger My solution allows only to specify whether to have short/long time/date. @mosberger s solution does allow custom formatting.

Lets discuss about whether or not we should provide a custom string format on TimePicker/DateTimePicker. My opinion is to have a behavior like MS provided it in DatePicker; to not let the user change the string Format like he wants to. Why do we need this bahavior. For each culture the date-time format is already predefined since ages.

PR #2965, opened 2017-05-11
Added ComboDataGrid Control

zsKengren:
## What changed?
Added the new Combo DataGrid Control. 

The ComboDataGrid acts just like the ComboBox but instead of ComboBoxItems, it has a DataGrid to display the data.

The code was migrated from the Elysium Extra project.

image


zsKengren:
I am sorry I didn't clean up my code better. I may have gotten a little too excited to get this up.

punker76:
Thanks for this, and sorry that it has been sitting here for so long without an answer.

I am closing it, and not because of the idea. The branch is from 2017 and written against the layout we had then, the NET40 and NET45 project pairs and the shared projitems, none of which exist any more. Bringing it forward would mean writing it again against today's sources, so as a pull request it has run out of road.

The idea itself stays open in #4033, and your template is the best starting point anyone has for it. One thing I would do differently, having just measured it: deriving from `ComboBox` does not work out as well as it looks. Without an `ItemsPresenter` the combo box generates no `ComboBoxItem` containers at all, so `ItemContainerStyle`, `ItemTemplate` and the item keyboard handling are inherited and dead while staying visible on the control. Both also share the same `CollectionView`, so the grid's new item placeholder turns up in the combo box's own `Items` and its `SelectedIndex`. And in the automation tree the combo box reports no contents, because the popup lives in a tree of its own and only the real containers would have carried the items across.

A control of its own, with a popup and a grid inside, keeps the public surface honest and avoids all three.

Thanks again for the work back then, and sorry it took this long to say so.

PR #2971, opened 2017-05-20
Added event delegates to the show dialog methods

MurariAlex:
## What changed?

The new event delegates provides extra flexibility to the dialog coordinator class.
If you want, for example, always close all open dialogs before showing another dialog, you can create an method and add it to the delegate that is invoked before showing the dialog. The current solution for this is to replicate code or wrap the the methods in another method.

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Dialogs/DialogCoordinator.cs
+++ b/src/MahApps.Metro/MahApps.Metro.Shared/Controls/Dialogs/DialogCoordinator.cs
@@ -13,54 +13,72 @@ namespace MahApps.Metro.Controls.Dialogs
 
         public Task<string> ShowInputAsync(object context, string title, string message, MetroDialogSettings settings = null)
         {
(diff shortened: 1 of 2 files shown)


MurariAlex:
Turns out that synchronous events and asynchronous methods aren't best friends, so I'll work on another solution for this subject.

bigworld12:
instead of using events you can use List<Action> which helps you control the execution in threads
and then wrap custom events around these lists for easy add/remove
then internally you just loop through the list and execute each action in an asynchronous pattern 
check this
https://stackoverflow.com/questions/2434187/action-t-synchronous-and-asynchronous

thoemmi:
IMHO this is a very special requirement.

> If you want, for example, always close all open dialogs before showing another dialog, you can create an method and add it to the delegate that is invoked before showing the dialog.

You would have to track all open dialogs too. Probably you want to close not all but only certain dialogs. Introducing the events in the PR wouldn't help much. I think this belongs to the business logic of your app, and not into the presentation layer.

PR #2991, opened 2017-06-22
Enable LayoutTransform and RenderTransform on MetroWindow

batzen:
This PR enables LayoutTransform and RenderTransform on MetroWindow by forwarding LayoutTransform and RenderTransform to the first template child.


batzen:
Strange, this compiles on my machine and i only added two property forwards in MetroWindow. Don't get why this did not compile on TeamCity. Any ideas?

punker76:
👍

PR #2992, opened 2017-06-22
Fixing a few issues in ThemeManager

batzen:
- ThemeManager did not look at merged dictionaries the way wpf did. wpf looks at them in a reverse order and so should we.
- StringComparison should be OrdinalIgnoreCase
- Switched to nameof instead of hard coded values
- Switched to string interpolation instead of string.format
- Switched to null conditional operator instead of `SafeRaise`
- Added some missing documentation
- Removed "metro" from documentation to ease sharing of code with Fluent.Ribbon


punker76:
@batzen 👍

PR #2997, opened 2017-06-27
[Fix] WPF-BindingLeak: Bind to property of non INotifyPropertyChanged without using BindingMode.OneTime

xxMUROxx:


punker76:
@xxMUROxx 👍

thoemmi:
@xxMUROxx 

I have one objection: `ThicknessToDoubleConverter.ConvertBack` was changed to throw an exception instead of returning `DependencyProperty.UnsetValue`. What's the reason for that change? The [documentation for `IValueConverter.ConvertBack`](https://msdn.microsoft.com/en-us/library/system.windows.data.ivalueconverter.convertback(v=vs.110).aspx) says (emphasis mine):

> The data binding engine does not catch exceptions that are thrown by a user-supplied converter. **Any exception that is thrown by the ConvertBack method, or any uncaught exceptions that are thrown by methods that the ConvertBack method calls, are treated as run-time errors. Handle anticipated problems by returning `DependencyProperty.UnsetValue`**.
>
> A return value of `DependencyProperty.UnsetValue` indicates that the converter produced no value and that the binding uses the FallbackValue, if available, or the default value instead.
>
> A return value of `Binding.DoNothing` indicates that the binding does not transfer the value or use the `FallbackValue` or default value.

/cc @punker76

punker76:
@thoemmi Ah, yeah, your're right for this. I'll change it.

xxMUROxx:
@thoemmi thanks for the notification. It was not intended to change this piece of code.

PR #3002, opened 2017-07-04
[Fix] DateTime is shown as "07.04.2017 10:32 A7" instead of "07.04.2017 10:32 AM"

xxMUROxx:


punker76:
@xxMUROxx it's already fixed

xxMUROxx:
@punker76 sorry, was not on the updated develop branch.

PR #3019, opened 2017-08-10
Fix a typo that stops users from setting the CloseButtonMargin proper…

5pyd3r:
…ty of MetroTabItem.

The fix is really simple and straightforward.

        public Thickness CloseButtonMargin
        {
             get { return (Thickness)GetValue(CloseButtonMarginProperty); }
    -        set { SetValue(CloseButtonEnabledProperty, value); }
    +        set { SetValue(CloseButtonMarginProperty, value); }
        }


punker76:
@5pyd3r 👍

PR #3032, opened 2017-08-31
Add DisabledVisualElementVisibility property for buttons

ye4241:
## What changed?

Add DisabledVisualElementVisibility property for buttons( button and togglebutton), it looks good when in dark theme.

**Before:** too highlight for disabled button
image

**After:** looks good
image


ye4241:
add font weight support for toggle switch.

PR #3035, opened 2017-09-05
progress bar has been added to screen content

eliezeralmeida:
I added a progress bar to MetroWindow, which can be activated using the Working property.


punker76:
@eliezeralmeida Thx for contributing, but I think such a feature should not be added to this lib, which is in main case a style/theme lib. You can add this feature if you need this in your own app by adding this to you rmain content and you will get the same behavior.

/cc @thoemmi

thoemmi:
I agree with @punker76, that's application logic. Every user of MahApps.Metro can add a spinner to his window, it's not a styling concern.

eliezeralmeida:
A spinner is something useful for a better UX. Adding it to all windows is a boring job, you have the option to create a new window by extending the MahApp.Window, so the style should checked and updated every MahApp update. So the proposal to add it directly in MahApp. Thank you very much for the return.

thoemmi:
@eliezeralmeida Your request is reasonable, but having a spinner on every window is a very special requirement. Many users of MahApps.Metro have their own requirements, each of those reasonable for his use cases, but adding all those feature requests to the library whould lead to a feature explosion and would be hardly maintainable.

eliezeralmeida:
@thoemmi I understand

PR #3048, opened 2017-09-09
Tab Control Underline Position

jammerxd:
## What changed?

A dependency property was added to the tabcontrolhelper class that allows a person to set the underline position in the various tab controls. The property is a Dock type and can be set to Left,Right,Top, or Bottom respectively. Not very well tested but it was tested and appeared to be functioning as it should. Would like a second person to verify if possible.


punker76:
@jammerxd your changes have been merged, thanks for your contribution 👍

jammerxd:
Wow! I didn't know this was something that would make it in!

PR #3050, opened 2017-09-09
Allow TimePicker to parse times with an AM/PM.

toddburnside:
## What Changed?
_Changed TimePickerBase.OnTextBoxLostFocus() to parse the DatePickerTextBox text as a DateTime rather than a TimeSpan so that time strings with an AM/PM will be parsed correctly_

_Closed issues. #3049_

I would have added a test, but there are currently no tests for editing the time via the textbox, and it seemed non-trivial to add.


punker76:
@toddburnside thx! 👍

PR #3066, opened 2017-10-06
Don't catch TextChangedEvent in NumericUpDown

Silv3rcircl3:
## What changed?

The `NumericUpDown` control catches the `TextBoxChangedEvent` and therefore other styles that depend on the event doesn't work correctly, i.e. the `SmartHint` feature from Material Design.
Example: 

This PR removes the line which catches the event.


punker76:
@Silv3rcircl3 👍

Issue #2384, opened 2016-03-01
MahApps.Metro consistent style and template names/keys


punker76:
MahApps.Metro styles and template keys are non consostent.
### Expected outcome

Consistent names/keys for all styles and templates.

e.g.

`Win10WindowButtonCommandsTemplate` -> `MahApps.Metro.Templates.WindowButtonCommands.Win10`
`MetroFlatButton` -> `MahApps.Metro.Styles.FlatButton`

etc

amkuchta:
This is going to be so tedious :no_mouth: :no_mouth:

punker76:
@amkuchta v2.0 will be a great release -:-D

Evangelink:
Hi @punker76,

Is this ticket still active? If so, which naming do you want to go for? 

I'd be happy to give some help.

Evangelink:
@punker76 what do you think about the following convention:

`MahApps.Metro.` 
\+ `Style.` or `Template.` 
\+ Category (e.g. `Metro`, `Flat`, `Square`, `Win10`...) 
\+ Control (e.g. `Button`, `ToggleButton`...) 
\+ optional specifier (e.g. `Light`, `Dark`, `Accented`...)

which would give the following names:

- `MahApps.Metro.Style.Flat.Button`
- `MahApps.Metro.Style.Metro.Button`
- `MahApps.Metro.Style.Metro.WindowButton.Base`
- `MahApps.Metro.Style.Metro.WindowButton.Light`
- `MahApps.Metro.Style.Metro.TextBox.Search`
- `MahApps.Metro.Style.Metro.TextBox.Button`

Remarks:

- We could omit the `Metro` after `Style.`/`Template.` to avoid the redundancy but it would be less of a standard construction

- We could omit the `Metro` after `MahApps.` but I don't know from a product name POV if that's ok.

- We could merge the Control and specifier parts to get something like `SearchTextBox`, `ButtonTextBox`, `LightWindowButton`, `DarkWindowButton`...

WDYT?

punker76:
@Evangelink You are going in the same direction as I want to go:

I already changed some of the styles with these naming, like

- `MahApps.Metro.Styles.xyz`
- `MahApps.Metro.Styles.DataGrid.xyz` ... `xyz.Win10`
- `MahApps.Metro.Styles.Slider.xyz` ... `xyz.Win10`
- `MahApps.Metro.Styles.FlatSlider.xyz` ...

and

[...]

PR #3077, opened 2017-10-11
Add new MetroHeader control

punker76:
The `MetroHeaderControl` can contains any other content and can display a header on top of this content.

So you can do something like this:

with `HeaderTemplate`

```xaml
<Controls:MetroHeader Margin="2" Header="TextBox Header">
    <Controls:MetroHeader.HeaderTemplate>
        <DataTemplate>
            <StackPanel VerticalAlignment="Center" Orientation="Horizontal">
                <iconPacks:PackIconMaterial VerticalAlignment="Center" Kind="FaceProfile" />
                <TextBlock Margin="2 0 0 0"
                           VerticalAlignment="Center"
                           Text="{Binding}" />
            </StackPanel>
        </DataTemplate>
    </Controls:MetroHeader.HeaderTemplate>
    <TextBox Text="Enabled" />
</Controls:MetroHeader>
```

2017-10-11_15h58_35

or

2017-10-11_15h50_42

2017-10-11_16h04_32

This PR (control) is ispired by #1825 Header, headers, headers

Closes #1825

### DIFF
--- a/docs/release-notes/1.6.0.md
+++ b/docs/release-notes/1.6.0.md
@@ -47,6 +47,7 @@
 	+ `TabControlHelper.CloseTabCommand` attached property which executes if the TabItem will be closed.
 	+ `TabControlHelper.CloseTabCommandParameter` attached property which will be passed to the CloseTabCommand.
 - Add `TextBoxHelper.ButtonsAlignment` support for `PasswordBox` styles [#3074](https://github.com/MahApps/MahApps.Metro/pull/3074) [@waleswood](https://github.com/waleswood)
+- New `MetroHeaderControl` control which can contains any other content and can display a header on top of this content.
 
 ## Breaking Change
 
@@ -108,3 +109,4 @@ More informations about the reason of this decision can be found here:
 - [#957](https://github.com/MahApps/MahApps.Metro/issues/957) Expander icon in VS theme
 - [#1731](https://github.com/MahApps/MahApps.Metro/issues/1731) VS Theme TabItem Question
 - [#3009](https://github.com/MahApps/MahApps.Metro/issues/3009) Floating Watermak field doesn't fit according FontSize
+- [#1825](https://github.com/MahApps/MahApps.Metro/pull/1825) [WIP] Header, headers, headers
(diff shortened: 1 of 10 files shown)


amkuchta:
*heavy breathing*

punker76:
@thoemmi You are the naming specialist ;-)

`MetroHeaderControl` or `MetroHeader` ?

What do you think?

amkuchta:
@punker76 I'm not @thoemmi , but to me, since items like `Badged` don't have "control" in the name, this shouldn't either for consistency

amkuchta:
@punker76, looking through the controls that have already been designed / implemented, there seems to be a blend of those containing "Control" in the name (example: `MetroAnimatedTabControl`) and those that do not (example: `DateTimePicker`)  Perhaps, related to #2384, this might be a good time to nail down whether or not "Control" should be included throughout the library.  My vote is no, simply because Microsoft doesn't follow that convention - they named it a `TextBox`, not a `TextBoxControl`.  In my opinion, we should follow their naming template for consistency.

Obviously, my input doesn't amount to much at the end of the day - it's totally up to you and @thoemmi! :smile:

MurariAlex:
Agree with @amkuchta, besides, it´s the namespace that tells that an certain component is an control.

`MetroAnimatedTabControl` is inside the namespace `MahApps.Metro.Controls` giving the full qualified name `MahApps.Metro.Controls.MetroAnimatedTabControl`, which is pretty redundant.

Just `MahApps.Metro.Controls.MetroAnimatedTab` would be all right.

amkuchta:
@MurariAlex I ended up changing my example... Microsoft calls it a `TabControl` as well :laughing:

punker76:
@MurariAlex @amkuchta `MetroAnimatedTabControl` is a `TabControl`, so it should named xyzTabControl.

punker76:
@MurariAlex @amkuchta The `FlyoutsControl` controls the `Flyouts`, so it should names xyzControl.

@thoemmi I'll update the name from `MetroHeaderControl` to `MetroHeader`. It's derived from `GroupBox`, so we should give it also a name without the Control suffix.

amkuchta:
@punker76 the `FlyoutControl` makes sense. I just think that even Microsoft has it wrong - it should be something like `TabController` or `FlyoutController`.

But we aren't here to discuss/fix that particular thought :smile:

thoemmi:
Thanks for this entertaining discussion :lol:

First I wanted to suggest to derive from `GroupBox` instead of `HeaderedContentControl`, but you already changed that 😉 And I agree on naming it `MetroHeader`, without `Control`. It's like `GroupBox`, `TextBox`, you name it.

:shipit:

punker76:
@thoemmi

amkuchta:
piccardapproves

@punker76, any plans to create another pre-release so we can try out the headers?

punker76:
@amkuchta You can eat the always the latest nuget pre-release from this appveyor feed: https://ci.appveyor.com/nuget/mahapps.metro without any guaranty...

2017-10-12_21h44_50

2017-10-12_21h44_31

PR #3162, opened 2018-01-18
WIP MetroWindow: OverlayFadeIn / OverlayFadeOut

### LINKED ISSUE #2825: Shorten Dialog Show/Hide animation duration
I would like to shorten the Dialog Show/Hide animation duration significantly while not resorting to <code>MetroDialogSettings.AnimateHide = false</code>. What is the recommending approach to accomplish this?

### LINKED ISSUE #3022: HideMetroDialogAsync sometimes fails to close dialog

Attempting to call HideMetroDialogAsync and ShowMetroDialogAsync on two dialogs in rapid succession, with animations on, will sometimes cause HideMetroDialogAsync to never return and thus leave a dialog on the screen perpetually. This is a timing-based bug and can be somewhat difficult to reproduce.

I have done some debugging and have tracked the problem to an issue with WPF's Storyboard class; apparently it does not always fire its Completed event when multiple storyboards are created. Since BaseMetroDialog._WaitForCloseAsync listens to this event, it sometimes fails to complete too.

Note that, while this issue is similar to [#2753](https://github.com/MahApps/MahApps.Metro/issues/2753), the workaround suggested there is not sufficient to solve this problem; both AnimateShow and AnimateHide must be set to to false to avoid it. A possible workaround would be to timeout the _WaitForCloseAsync task if the Completed event takes too long to fire.

### Expected outcome

Dialogs attempting to open/close at the same time should do so successfully.

### Repo

This repo (originally submitted as a .zip file in issue [#2753](https://github.com/MahApps/MahApps.Metro/issues/2753) and submitted here with the original author's permission) can somewhat reliably reproduce the bug. The application opens a dialog at the press of a button; when that dialog is closed, a second dialog is very briefly opened and closed too. This will sometimes cause the original dialog to remain open. Repeatedly opening and closing the dialog should eventually trigger the bug.

https://github.com/AndreasZChen/MahAppsRace

### Environment

- MahApps.Metro __v1.5.0__
- Windows OS __10__
- Visual Studio __2017__
- .NET Framework __4.7__

punker76:
## What changed?

-  New `OverlayFadeIn` and `OverlayFadeOut` properties to set/override the overlay fade in/out Storyboards which are used for the Dialogs
- Fix the fade in and out animation if no duration exists
- Fix `HideMetroDialogAsync` which sometimes fails to close dialog (async call order)

_Closed issues._

Closes #2825 
Closes #3022

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.Shared/Controls/MetroWindow.cs
+++ b/src/MahApps.Metro/MahApps.Metro.Shared/Controls/MetroWindow.cs
@@ -789,6 +789,30 @@ namespace MahApps.Metro.Controls
             get { return TitleCaps ? Title.ToUpper() : Title; }
         }
 
+        private bool CanUseOverlayFadingStoryboard(Storyboard sb, out DoubleAnimation animation)
+        {
+            animation = null;
+            if (null == sb)
+            {
+                return false;
+            }
+
+            sb.Dispatcher.VerifyAccess();
+
+            animation = sb.Children.OfType<DoubleAnimation>().FirstOrDefault();
+            if (null == animation)
+            {
+                return false;
+            }
+
+            return (sb.Duration.HasTimeSpan && sb.Duration.TimeSpan.Ticks > 0)
+                   || (sb.AccelerationRatio > 0)
+                   || (sb.DecelerationRatio > 0)
+                   || (animation.Duration.HasTimeSpan && animation.Duration.TimeSpan.Ticks > 0)
+                   || animation.AccelerationRatio > 0
+                   || animation.DecelerationRatio > 0;
+        }
+
         /// <summary>
         /// Begins to show the MetroWindow's overlay effect.
         /// </summary>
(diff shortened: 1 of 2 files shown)


GreatFireWall:
Hi @punker76 , is there an example for using `OverlayFadeIn` and `OverlayFadeOut` ?
I want to make the `CustomDialog` show animation shorter but not disabled.

punker76:
@GreatFireWall `OverlayFadeIn` and `OverlayFadeOut` are properties at the `MetroWindow`, so simply set a new Storyboard and it will use it.

Here are the 2 standard Storyboards https://github.com/MahApps/MahApps.Metro/blob/3b72bcef9e964bf4ed93aa1a84683004ccd5fc69/src/MahApps.Metro/Themes/MetroWindow.xaml#L543-L556

GreatFireWall:
@punker76 thanks.

GreatFireWall:
I had try change the `AccelerationRatio` and `SpeedRatio` but no effect.
So should I add a `Duration` attribute ?

punker76:
@GreatFireWall And you did set the new Storyboards to the both propetries?

GreatFireWall:
Sorry my fault, now works well. :)

Issue #3238, opened 2018-05-04
FlatSlider height not working


hudec117:
## What steps will reproduce this issue?

Setting the `Height` property on a `Slider` with the `FlatSlider` style changes the height of the control but not of the slider itself.

In the following example, I set the `Height` to 30. As you can see the slider keeps the same height, but the designer shows that the control has grown to 30 pixels in height.

![1.6.2 example](https://raw.githubusercontent.com/hudec117/MahApps.Metro-Flat-Slider-Issue/master/Designer%201.6.2.PNG)

### Expected outcome

I expect the slider to fill the height of the control that it is in.

This wasn't a problem in __v1.5.0__. Here is an example of the same exact setup as above but using the __v.1.5.0__ package instead.

![1.5.0 example](https://raw.githubusercontent.com/hudec117/MahApps.Metro-Flat-Slider-Issue/master/Designer%201.5.0.PNG)

As you can see, the slider fills the height of the control.

### Repo

https://github.com/hudec117/MahApps.Metro-Flat-Slider-Issue

### Environment

- MahApps.Metro __v1.6.2__
- Windows OS __10__
- Visual Studio __2017__
- .NET Framework __4.7__

hudec117:
I don't know if it's intentional but the height of the track is set to the minimum height of the slider here:
https://github.com/MahApps/MahApps.Metro/blob/47c34962fc96b465092ae1a5b31e201c7a5fe530/src/MahApps.Metro/Styles/FlatSlider.xaml#L87

Given this, a work-around can be achieved by setting both the `MinHeight` and `Height` properties on the `Slider` to the same value. This will produce the same result as __v1.5.0__.

punker76:
@seb30 uups

seb30:
Did you plan to change this? Or is this more a feature than a bug?

punker76:
@seb30 oh, it's already fixed #3248 and released in [v1.6.4](https://github.com/MahApps/MahApps.Metro/releases/tag/1.6.4)

seb30:
@punker76 Thank you!

PR #3240, opened 2018-05-05
Fix for Track's height in FlatSlider style

hudec117:
## What changed?

Changed binding for the `Height` property in the Horizontal FlatSlider style to bind to the Slider's `Height` property instead of `MinHeight`.

_Closed issues_

Closes https://github.com/MahApps/MahApps.Metro/issues/3238


punker76:
@hudec117 thx for contributing. I changed your solution a little bit, so now it's possible to set a different height or width and control it additionally with the vertical and horizontal content property.

PR #3256, opened 2018-05-30
[3215] NumericUpDown - Hexdecimal input mode

Stephen-Ross:
I have removed the handling for non-digits within the preview text input and this is now deferred until we attempt to parse the text value into the numeric representation.

The user is now able to specify their own parsing number style instead of replying on `NumberStyles.Any`. This means that we can correctly parse the value based on what the user wants, this also means that we will parse the value as an integer if the user supplies an input mode of 'Number'. The default of this value is `NumerStyles.Any` so as to maintain backwards compatibility.

I've manually tested this with as many use cases and data as I could think of and the existing behaviour of the NumericUpDown appears to have been maintained.

There does not appear to be any unit tests covering the NumericUpDown so I was unable to validate against any expect behaviour that I may have manually missed.

### DIFF
--- a/src/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/src/MahApps.Metro/Controls/NumericUpDown.cs
@@ -141,6 +141,12 @@ namespace MahApps.Metro.Controls
             typeof(NumericUpDown),
             new PropertyMetadata(default(bool), OnSnapToMultipleOfIntervalChanged));
 
+        public static readonly DependencyProperty ParsingNumberStyleProperty = DependencyProperty.Register(
+            @"ParsingNumberStyle", 
+            typeof(NumberStyles),
+            typeof(NumericUpDown),
+            new PropertyMetadata(NumberStyles.Any));
+
         private static void IsReadOnlyPropertyChangedCallback(DependencyObject dependencyObject, DependencyPropertyChangedEventArgs e)
         {
             if (e.OldValue != e.NewValue && e.NewValue != null)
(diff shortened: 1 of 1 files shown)


punker76:
@Stephen-Ross thx for contributing with this feature. I changed your initial thought of separate checking decimal and numeric input, cause your changes doesn't works. A hexadecimal input can not be converted with double. The other change was, that the user now could type any character in. I think the current one is now nearly the same as before including your new property which allows now hexadicaml input. I will make some tests after this PR. thx

sensitron:
@Stephen-Ross When changing the StringFormat with a Trigger it will change the NumericInputeMode to all:

        private static void OnStringFormatChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
        {
            NumericUpDown nud = (NumericUpDown)d;

            if (nud._valueTextBox != null && nud.Value.HasValue)
            {
                nud.InternalSetText(nud.Value);
            }

            var value = (string)e.NewValue;

            if (!nud.NumericInputMode.HasFlag(NumericInput.Decimal) && !string.IsNullOrEmpty(value) && RegexStringFormatHexadecimal.IsMatch(value))
            {
                nud.SetCurrentValue(NumericInputModeProperty, nud.NumericInputMode | NumericInput.Decimal);
            }
        }

Which will result in an Exception when typing a hexadecimal value into the TextBox.

        private bool ValidateText(string text, out double convertedValue)
        {
            // If we are only accepting numbers then attempt to parse as an integer.
            if (NumericInputMode == NumericInput.Numbers)
            {
                return ConvertNumber(text, out convertedValue);
            }

            if (!double.TryParse(text, ParsingNumberStyle, SpecificCultureInfo, out convertedValue))
            { ......

=> System.ArgumentException, because AllowHexSpecifier can't be parsed.

My current workaround:

Also set the NumericInputeMode to Numbers in the Trigger:

                        <controls:NumericUpDown.Style>
                            <Style TargetType="controls:NumericUpDown" BasedOn="{StaticResource {x:Type controls:NumericUpDown}}">
                                <Setter Property="NumericInputMode" Value="Numbers"></Setter>
                                <Setter Property="ParsingNumberStyle" Value="Any"></Setter>
                                <Setter Property="StringFormat" Value=""></Setter>
                                <Style.Triggers>
                                    <DataTrigger Binding="{Binding ElementName=HexaCheckBox, Path=IsChecked}" Value="True">
                                        <Setter Property="ParsingNumberStyle" Value="AllowHexSpecifier"></Setter>
                                        <Setter Property="StringFormat" Value="{}{0:X0}"></Setter>

                                        <Setter Property="NumericInputMode" Value="Numbers"></Setter>
                                    </DataTrigger>
                                </Style.Triggers>
                            </Style>
                        </controls:NumericUpDown.Style>

sensitron:
Oh nevermind @punker76 ... I saw you already fixed that problem in https://github.com/MahApps/MahApps.Metro/pull/3268, sorry :P

PR #3268, opened 2018-06-14
(GH-3189) Fix NumericUpDown Input Issues with StringFormat

### LINKED ISSUE #3189: NumericUpDown Input Issues
_________________________
Issue 1 

Example: NumericUpDown with StringFormat="{}{0:N2} cm"
Initial Value be "100.00 cm". 
Input "200.00 cm" -> working fine
Input "200.00cm" (Without space) -> value is not accepted, gets back to "100.00 cm"
Input "200.00" -> working fine, goes to "200.00 cm"
Input "200.00m" -> not accepted, gets back to "100.00 cm"

Makes it very inconvenient to type values, especially in combination with Issue 2:

________________________
Issue 2

Example: NumericUpDown with StringFormat="{}{0:N2} cm"
Initial Value be "100.00 cm". 

Doubleclick into NumericUpDown should mark the whole text ("100.00 cm") or at least the value without space ("100.00"). This way you could input values conveniently even with Issue 1 still existing.
At the moment if you doubleclick it will mark either before or after decimal dot ("100" or "00").

### Environment

- MahApps.Metro 1.6.0.184
- Windows OS 7 64bit
- Visual Studio 2017 community
- .NET Framework 4.6.1

### LINKED ISSUE #3217: NumericUpDown Parses Decimal Value Incorrectly
The NumericUpDown control is improperly parsing a decimal value that is entered with no leading zero in the en-US culture. For example, .1 becomes 1.0, 0.105 becomes 105, etc.

This is a really easy issue to reproduce. An entire working example showing the failure is:
```
    class Program
    {
        // The regex copy pasted out of numericupdown control file...
        private static readonly Regex RegexStringFormatNumber = new Regex(@"[-+]?(?<![0-9][.,])\b[0-9]+(?:[.,\s][0-9]+)*[.,]?[0-9]?(?:[eE][-+]?[0-9]+)?\b(?!\.[0-9])", RegexOptions.Compiled);

        static void Main(string[] args)
        {
            // returns 1
            string leadingdec = GetAnyNumberFromText(".1");
            
            // returns 0.1
            string leadingZero = GetAnyNumberFromText("0.1");

            // returns 0.1 in double
            double.TryParse(".1", out double leadingDecDouble);

            // also returns 0.1 in double
            double.TryParse("0.1", out double leadingZeroDouble);
        }

        // The GetAnyNumberFromText function copy pasted out of numericupdown control file...
        private static string GetAnyNumberFromText(string text)
        {
            var matches = RegexStringFormatNumber.Matches(text);
            if (matches.Count > 0)
            {
                return matches[0].Value;
            }
            return text;
        }
    }
```

### Expected outcome

We expect the parsed value to match the value that double.Parse would return on the raw string. For this example, ".1" should parse to 0.1 not 1.0 which it does currently.

### Environment

- MahApps.Metro: Latest master has issue
- Windows OS 10
- Visual Studio 2017
- .NET Framework 4.5

### LINKED ISSUE #3224: NumericUpDown ignores value change with stringformat if only the number is selected
Select value and enter text by hand will be completely ignored if stringformat is used and the user only selects the value but not everything.

Example:
`StringFormat=" {0} mb "`

If user not selects 'mb' as well, the value will be ignored. This is problematically since the focus on double click is automatically set only to the value.

Attached example project. Launch it and double click on the 5000 and enter a new value. Click the button. Next double click on the 5000 and press CTRL+A and then enter a new value.

[WpfApp2.zip](https://github.com/MahApps/MahApps.Metro/files/1942488/WpfApp2.zip)

### Environment

- MahApps.Metro 1.6.1
- Visual Studio 2017
- .NET Framework 4.7.1

punker76:

Fix NumericUpDown Input Issues with StringFormat and parsing numbers.

**Unit test**

*NumericUpDownTests* 

**Closed Issues**

Closes #3189 
Closes #3217 
Closes #3224

### DIFF
--- a/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/TextExamples.xaml
+++ b/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/TextExamples.xaml
@@ -368,7 +368,7 @@
                                         IsReadOnly="{Binding ElementName=ReadOnlyCheck, Path=IsChecked, Mode=TwoWay}"
                                         NumericInputMode="{Binding ElementName=NumericInputCheckBox, Path=SelectedItem, Mode=TwoWay}"
                                         StringFormat="0,000.00"
-                                        Value="{Binding ElementName=test, Path=Value, Mode=TwoWay}" />
+                                        Value="5" />
 
                 <Label Content="Select all on focus" />
                 <Controls:NumericUpDown Controls:TextBoxHelper.SelectAllOnFocus="True"
--- a/src/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/src/MahApps.Metro/Controls/NumericUpDown.cs
@@ -142,7 +142,7 @@ namespace MahApps.Metro.Controls
             new PropertyMetadata(default(bool), OnSnapToMultipleOfIntervalChanged));
 
         public static readonly DependencyProperty ParsingNumberStyleProperty = DependencyProperty.Register(
-            @"ParsingNumberStyle", 
+            "ParsingNumberStyle", 
             typeof(NumberStyles),
             typeof(NumericUpDown),
             new PropertyMetadata(NumberStyles.Any));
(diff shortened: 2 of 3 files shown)


DanHickstein:
These are really nice features! (#3217 is especially useful for us.)

As of Oct 2019 this is still not incorporated into a released version - can we expect v2.0 soon?

Thanks for your great work!

Issue #3293, opened 2018-07-27
IsEnabled does not work properly on HamburgerMenuItem


igorquintaes:
**Bug Description**
Is not possible to enable or disable a _HamburgerMenuItem_ inside a .xaml file. 
If a *HamburgerMenuItem* contains an _IsEnabled_ attribute, does not matter if it is True or False, binded value or not - it always shows as disabled.

**To Reproduce**
Based on [punker76/code-samples](https://github.com/punker76/code-samples):
1. Open HamburgerMenuApp.V2
2. Find a _HamburgerMenuGlyphItem_ in MainWindow.xaml
3. Add _IsEnabled="True"_

**Expected behavior**
Menu item should still be enabled.

**Environment(please complete the following information):**
Based on [punker76/code-samples](https://github.com/punker76/code-samples):
- OS: Win10 1803
- VS 2017 15.7.3

Based on my own code:
- MahApps.Metro updated
- OS: Win10 1803
- VS 2017 15.7.3
- .NET Framework 4.5

**Repo**
Since it can be simulated on [punker76/code-samples](https://github.com/punker76/code-samples), i think it is not needed.

**Additional context**
Based on Issue #2865, and Pull Request #2921 it is expected to work.

realivanjx:
same here. i changed IsEnabled via C# code but if i disabled a menuitem then i cant enable it again using IsEnabled = True :(

Anapher:
I don't think that this is solved as I just cloned the current develop branch and the problem is still there. I'm pretty sure that the problem is caused [by this line](https://github.com/MahApps/MahApps.Metro/blob/develop/src/MahApps.Metro/Controls/HamburgerMenu/MenuItems/HamburgerMenuItem.cs#L261). In this [void](https://github.com/MahApps/MahApps.Metro/blob/develop/src/MahApps.Metro/Controls/HamburgerMenu/MenuItems/HamburgerMenuItem.cs#L243), `CanExecute` is set to `true` if the `Command` is `null`. The problem is, that on creation the `Command` is `null` but `CanExecute` is `false` which is a wrong state. I will submit a PR that changes [this line](https://github.com/MahApps/MahApps.Metro/blob/develop/src/MahApps.Metro/Controls/HamburgerMenu/MenuItems/HamburgerMenuItem.cs#L264) to ` private bool canExecute = true;`

Hanebu:
[...]

PR #3309, opened 2018-08-18
Fix HamburgerMenu IsEnabled false by default

Anapher:
The problem occurred when setting the property 'IsEnabled' to true, the item became disabled. The problem was the `CanExecute` property should be true if the `Command` is _null_ (as specified in `UpdateCanExecute()`), but at the initialization of the control, the local variable `canExecute` has the default value _false_ even if the `Command` is _null_. That leads to the `IsEnabledCoerceValueCallback` returning _false_ by default, so if the `IsEnabled` property is set to any value (so the callback is triggered), it defaults to _false_.

**Closes Issues**
#3293


punker76:
@Anapher Thx for the fix 👍

PR #3404, opened 2018-12-13
single BooleanToVisibilityConverter naming consistency

swdriessen:
**Changes**

Renamed old BooleanToVisibilityConverter key for the sake of naming consistency.


punker76:
@swdriessen Thx

PR #3428, opened 2019-01-14
Added more choices for positioning buttons on NumericUpDown control

MarcoMartins86:
Added more choices for positioning buttons on NumericUpDown control namely:
- New **ButtonsAlignment="OpposingSides**" option that will basically have the text box in the middle of the buttons.
- New **SwitchUpDownButtons** property that will switch the buttons places

For testing, I have changed the NumericUpDown that controls the index of FlipView2nd in the OtherExamples.xaml file.

### DIFF
--- a/src/MahApps.Metro/Controls/ButtonsAlignment.cs
+++ b/src/MahApps.Metro/Controls/ButtonsAlignment.cs
@@ -3,6 +3,7 @@
    public enum ButtonsAlignment
    {
       Left,
-      Right
+      Right,
+      OpposingSides
    }
 }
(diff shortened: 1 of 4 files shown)


src/MahApps.Metro/Controls/ButtonsAlignment.cs
```diff
@@ -3,6 +3,7 @@
    public enum ButtonsAlignment
    {
       Left,
-      Right
+      Right,
+      OpposingSides
```
punker76: @MarcoMartins86 I think `Opposite` fits better with the other two values.

MarcoMartins86: Yes, you're right. Done.

punker76:
@MarcoMartins86 👍

PR #3475, opened 2019-03-19
Provide event for hamburger button click

floele-sp:
This adds a HamburgerButtonClick event to the HamburgerMenu control.

I needed this particular event for saving the UI settings, that is save the current "pane open" state when the button has been clicked, as opposed to other situations where IsPaneOpen may be changed programmatically.

I used the ItemClickEventArgs for this event with the hamburger button as value. I didn't need this particular implementation (without would have been fine as well) but it might be useful for other purposes and shouldn't hurt.

### DIFF
--- a/src/MahApps.Metro/Controls/HamburgerMenu/HamburgerMenu.Events.cs
+++ b/src/MahApps.Metro/Controls/HamburgerMenu/HamburgerMenu.Events.cs
@@ -24,9 +24,16 @@ namespace MahApps.Metro.Controls
         /// </summary>
         public event EventHandler<HamburgerMenuItemInvokedEventArgs> ItemInvoked;
 
-        private void HamburgerButton_Click(object sender, RoutedEventArgs e)
+        /// <summary>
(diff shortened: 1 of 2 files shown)


floele-sp:
@punker76 Is there something wrong with my pull-request? If not, I'd appreciate if you consider merging it. Thanks!

punker76:
Your changes have been merged, thanks for your contribution 👍

I changed the usage of the ItemCliickEventArgs, because it's not this thing. I use a RoutedEvent with RoutingStrategy.Bubble (with this we can react to the Handled property).

```xaml
private void HamburgerMenuControl_OnHamburgerButtonClick(object sender, RoutedEventArgs e)
{
    // we want to open/close the pane by us
    e.Handled = true;

    // do something special

    // now open/close the pane by us
    IsPaneOpen = !IsPaneOpen;
}
```

https://github.com/MahApps/MahApps.Metro/commit/e9fa9637a5ff76da98b823a37ebd7b891cd18c29

PR #3514, opened 2019-05-30
Use TabControlEx from ControlzEx as base class for BaseMetroTabControl

batzen:
This PR changes the base class of BaseMetroTabControl from TabControl to TabControlEx from ControlzEx.
This allows us to keep the visual content in memory and prevent reloads of the UI contained in Tabs.
To be as flexible as possible i have added a need property `KeepVisualTreeInMemoryWhenChangingTabs` to `MetroTabControl`.
The default value is currently set to `false` to not break any current usages of the control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ artifacts/
 *.tmp
 *.tmp_proj
 *.log
+*.binlog
 *.vspscc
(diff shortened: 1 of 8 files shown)


punker76:
@batzen 👍 🛩

PR #3522, opened 2019-06-05
Added a property to control value of NumericUpDown.

wunianqing:

Please see the demo application. I have added a small demo in it.
The original `NumericUpDown` control will response every enter. This will cause a strange behavior when you binding value to a control like Slider. So I added a property to allow value effect after press enter. 

**Unit test**

I don't know how to test a custom control

**Additional context**

None

**Closed Issues**

None


src/MahApps.Metro.sln
```diff
@@ -36,10 +36,10 @@ Global
...
 		{D9ACF0D0-A178-4C91-BD23-D013CEA763A0}.Release|Any CPU.Build.0 = Release|Any CPU
-		{40048BCF-D1C7-46CC-B781-05718BE25BFC}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{40048BCF-D1C7-46CC-B781-05718BE25BFC}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{40048BCF-D1C7-46CC-B781-05718BE25BFC}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{40048BCF-D1C7-46CC-B781-05718BE25BFC}.Release|Any CPU.Build.0 = Release|Any CPU
+		{31D47FA4-9A93-4385-A326-76F189C02440}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{31D47FA4-9A93-4385-A326-76F189C02440}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{31D47FA4-9A93-4385-A326-76F189C02440}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{31D47FA4-9A93-4385-A326-76F189C02440}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
```
punker76: Are these changes really necessary?

src/MahApps.Metro/MahApps.Metro.csproj
```diff
@@ -10,8 +10,8 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
     </PropertyGroup>
     <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.0'">
-        <Reference Include="System.ComponentModel.DataAnnotations"/>
-        <Reference Include="System.Configuration"/>
+        <Reference Include="System.ComponentModel.DataAnnotations" />
+        <Reference Include="System.Configuration" />
     </ItemGroup>
```
punker76: Are these changes really necessary?

src/MahApps.Metro/Styles/Controls.NumericUpDown.xaml
```diff
@@ -0,0 +1,20 @@
...
+           TargetType="{x:Type Controls:NumericUpDown}"
+           BasedOn="{StaticResource {x:Type Controls:NumericUpDown}}">
+        <Style.Resources>
+            <Style x:Key="ChromelessButtonStyle" TargetType="{x:Type ButtonBase}"
+                   BasedOn="{StaticResource ChromelessButtonStyle}">
+                <Setter Property="Focusable" Value="False"/>
+            </Style>
+        </Style.Resources>
+    </Style>
+</ResourceDictionary>
```
punker76: Why do we need a new style?

src/MahApps.Metro.sln
wunianqing: Not necessary. It was added by VS automatically.

src/MahApps.Metro/MahApps.Metro.csproj
wunianqing: No. I just pressed Ctrl+K, Ctrl+D

src/MahApps.Metro/Styles/Controls.NumericUpDown.xaml
wunianqing: Yes, we do need a new style. We need TextBox stay in focusing when click increase or decrease button. But we can make this shorter. I committed a new style in my repository.

punker76: > Yes, we do need a new style. We need TextBox stay in focusing when click increase or decrease button. But we can make this shorter. I committed a new style in my repository.

I think we can get the same result without creating/using a new style. We can prevent focusing of the up and down buttons permanently when we do this

image

image

wunianqing: Yes, of course. I created a new style just want to keep the original behavior. If you think this is OK. Then we don't need a new style any more.

wunianqing: Wait. I think I misunderstood you.

punker76: I will think about another DepProp, which will control this behavior and check into the PR branch here.

wunianqing: The purpose of my style is that when you set `ValueChangedOnLostFocus` to true, then you may want increase button and decrease button only effect the text, not the value. The value only changes when press enter.

punker76: No, that's not a good UX. The value should changed by using up and down buttons. Only if you type manually we should have this behavior.

wunianqing: Yes, but this only happens when you focused on TextBox. If you are not focusing TextBox, then the buttons work.

wunianqing: And that's why I created a new style rather than change it directly on the default style. It is just an option.

punker76: No new style needed. Same can be done with e.g. a DepPropery UpDownButtosFocusable with default true to keep the current implementation. Then another property to handle your new requirement to change the value behind the user input manually by hitting enter.

wunianqing: Cool. I am OK with that.

PR #3526, opened 2019-06-06
NumericUpDown: Add a property to control changing value with key Enter

punker76:

This PR was create after discussing with @wunianqing (#3522): The `NumericUpDown` control will response on every enter and change their value. This will cause a strange behavior when you binding value to a control like Slider. So I added a property to allow value effect after press enter.

- [x] Add a new dependency property `ChangeValueOnTextChanged` to enable changing value from manually user input. This property indicates whether the value will be changed directly on every TextBox text changed event or when using the Enter key.
- [x] Add also a new dependency property `UpDownButtonsFocusable` to allow changing the Focusable property of the up and down buttons.
The default is true to keep the  current behavior. If the up or down button will be used then the focus will move to it and triggers the lost focus event of the inner TextBox to change the value.

**Additional context**

**Closed Issues**


punker76:
@wunianqing your changes have been merged, thanks for your contribution 👍

PR #3546, opened 2019-07-09
[GH-3544] Add null check for Application.Current

### LINKED ISSUE #3544: Application.Current: Missing Null check
**Describe the bug**
`2.0.0-alpha0262`
You use often `Application.Current`. Is the XAML Window hosted in a Windows Forms application, it could be that `Application.Current` is null.

Here is a example stacktrace:
```
MahApps.Metro.Behaviors.WindowsSettingBehavior.AssociatedObject_SourceInitialized(Object sender, EventArgs e) in C:\projects\mahapps-metro\src\MahApps.Metro\Behaviors\WindowsSettingBehavior.cs:Zeile 45.
```

**To Reproduce**
Host a `MetroWindow` in a Windows Forms application. 

**Expected behavior**
No exception 🦄

Evangelink:

Add null checks for usage of `Application.Current` in the main part of the library. Note that the demo and the unit tests haven't been updated as they are unlikely to be used wrapped in a WinForm App.

**Unit test**

No unit test added but if needed I can add a new test project with a WinForm app hosting the WPF MahApps window.

**Additional context**

**Closed Issues**

Fix #3544

### DIFF
--- a/src/MahApps.Metro/ThemeManager/ThemeManager.cs
+++ b/src/MahApps.Metro/ThemeManager/ThemeManager.cs
@@ -374,10 +374,21 @@ namespace MahApps.Metro
         /// <returns>The resource object or null, if the resource wasn't found.</returns>
         public static object GetResourceFromAppStyle(Window window, string key)
         {
-            var appStyle = (window.IsNull()
+            Theme appStyle;
+
+            if (Application.Current.IsNull()) // In case the Window is hosted in a WinForm app
+            {
+                appStyle = !window.IsNull()
+                                ? DetectTheme(window)
+                                : null;
+            }
+            else
+            {
+                appStyle = (window.IsNull()
                                 ? DetectTheme(Application.Current)
                                 : DetectTheme(window))
-                           ?? DetectTheme(Application.Current);
+                            ?? DetectTheme(Application.Current);
+            }
 
             var resource = appStyle?.Resources[key];
 
@@ -786,8 +797,9 @@ namespace MahApps.Metro
                 }
             }
 
-            // ReSharper disable once AssignNullToNotNullAttribute
-            return DetectTheme(Application.Current);
+            return Application.Current != null // In case the Window is hosted in a WinForm app
+                ? DetectTheme(Application.Current)
+                : null;
         }
 
         /// <summary>
@@ -804,7 +816,9 @@ namespace MahApps.Metro
             }
 
             var detectedStyle = DetectTheme(window.Resources)
-                                ?? DetectTheme(Application.Current.Resources);
+                                ?? (Application.Current != null // In case the Window is hosted in a WinForm app
+                                        ? DetectTheme(Application.Current.Resources)
+                                        : null);
 
             return detectedStyle;
         }
@@ -937,7 +951,10 @@ namespace MahApps.Metro
                                ? BaseColorLight
                                : BaseColorDark;
 
-            ChangeThemeBaseColor(Application.Current, baseColor);
+            if (Application.Current != null) // In case the Window is hosted in a WinForm app
+            {
+                ChangeThemeBaseColor(Application.Current, baseColor);
+            }
         }
 
         private static bool isAutomaticWindowsAppModeSettingSyncEnabled;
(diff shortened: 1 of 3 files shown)


src/MahApps.Metro/ThemeManager/ThemeManager.cs
```diff
@@ -374,10 +374,21 @@ public static bool IsThemeDictionary([NotNull] ResourceDictionary resources)
         /// <returns>The resource object or null, if the resource wasn't found.</returns>
         public static object GetResourceFromAppStyle(Window window, string key)
         {
-            var appStyle = (window.IsNull()
+            Theme appStyle;
+
+            if (Application.Current.IsNull()) // In case the Window is hosted in a WinForm app
```
Evangelink: I have used the `IsNull()` extension method here as it seems to be used in this class. If needed I could extract the methods `IsNull()` and `IsNotNull()` into a separate class and use them for the other changes.

src/MahApps.Metro/ThemeManager/ThemeManager.cs
```diff
@@ -937,7 +951,10 @@ public static void SyncThemeWithWindowsAppModeSetting()
                                ? BaseColorLight
                                : BaseColorDark;
 
-            ChangeThemeBaseColor(Application.Current, baseColor);
+            if (Application.Current != null) // In case the Window is hosted in a WinForm app
```
Evangelink: I have simply avoided the case where the current app is null but we might want to throw or do something else in this case.

punker76: We can leave this as it is, because there is already a check at the [HandleUserPreferenceChanged](https://github.com/MahApps/MahApps.Metro/blob/fbc736f286022bf8f2086dbd5f0b16321714d777/src/MahApps.Metro/ThemeManager/ThemeManager.cs#L972-L986) method. The other call is only at the OnStartup of the App itself.

src/MahApps.Metro/ThemeManager/ThemeManager.cs
punker76: I don't know what is better in these cases. Should we always use these 2 extension methods? Or not? I don't know...

Evangelink: TBH I am not really sure to understand what those 2 methods bring compare to having the null check. Was it to avoid some FPs from ReSharper with the null reference rules?

punker76: Yes this could be the reason, @batzen can explain this a little bit more.

batzen: I added those because writing `if (!(xy is null))` feels horrible, whereas `if (xy.IsNotNull())` does not. And to be consistent i also added `IsNull`.

batzen:
Would you mind added your changes/fixes also to Fluent.Ribbon? 
We (@punker76 and i) try to keep the classes in sync.

punker76:
@Evangelink your changes have been merged, thanks for your contribution 👍

src/MahApps.Metro/ThemeManager/ThemeManager.cs
Evangelink: @batzen I don't know whether this is a good practice or not but you could use `if (x is object)` to test non-nullability.

Evangelink:
@batzen I'll do the PR on `FluentRibbon` today!

src/MahApps.Metro/ThemeManager/ThemeManager.cs
timunie: just to complete this: Since C# 7.0 you could also use this code:
```
        if (SomeObject is SomeType variableName)
        {
            // Do some stuff with variableName
            return variableName.ToString();
        }
```

More about this: [https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/is](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/is)

Happy coding 

Tim

punker76: @timunie That's correct, and I use it more and more, because of better readable code, but sometimes you don't need extra vars ;-)

PR #3572, opened 2019-08-11
Improve check for RD "equality"

batzen:
Only check RDs that have keys and only if first RD has at least as many keys as second.
Otherwise we might detect RDs as equal that have zero keys.
Checking if the first RD has at least as many keys as the second just ensures that we don't compare apples with pineapples ;-)


punker76:

Issue #2864, opened 2017-02-22
Visual Studio Theme Context Menu


defunky:
By creating a context menu with VS theme, the style applied seems to be broken Menu/Menu Item seem to be fine though

![image](https://cloud.githubusercontent.com/assets/17749501/23190915/ce307f72-f892-11e6-988e-791389129bab.png)

punker76:
@defunky Please, can you post a full repo on GutHub to investigate this issue? Thx

JobaDiniz:
@punker76 why this was closed?
The repo is simple, in the `VSDemo.xaml`, add a static `ContetMenu` and you'll see the style.

JobaDiniz:
Video of the error happening in the MahApps demo app:
![context-menu](https://user-images.githubusercontent.com/165290/52208807-0578f100-2869-11e9-8771-75394eab5412.gif)

**VSDemo.xaml** modified:
```
<Controls:MetroWindow x:Class="MetroDemo.ExampleWindows.VSDemo"
                      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                      xmlns:Controls="clr-namespace:MahApps.Metro.Controls;assembly=MahApps.Metro"
                      xmlns:System="clr-namespace:System;assembly=mscorlib"
                      Title="Visual Studio 2012 Style Demo"
                      Width="1000"
                      Height="800"
                      Style="{DynamicResource VSWindowStyleKey}"
                      WindowStartupLocation="CenterScreen">

    <Window.Resources>
        <ResourceDictionary>
            <ResourceDictionary.MergedDictionaries>
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/VS/Colors.xaml" />
                <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/VS/Styles.xaml" />
            </ResourceDictionary.MergedDictionaries>
        </ResourceDictionary>
    </Window.Resources>

    <DockPanel>
        <Menu Margin="2 0 0 0"
              DockPanel.Dock="Top"
              SnapsToDevicePixels="True">
            <MenuItem Header="ROCKETS">
                <MenuItem Header="Rocket slot 1">
                    <MenuItem Header="Fire slot" />
                    <Separator />
                    <MenuItem Header="Relaod slot" />
                    <MenuItem Header="Report damage to control room" />
[...]

PR #3624, opened 2019-10-03
Fix that makes ContextMenu work with VS theme

mgnslndh:
- Added resources to make the ContextMenu look correct when using the VS theme
- Made minor improvements to the default Button style for the VS theme

This PR is a fix for #2864 

ContextMenu

- Add VS Context Menu style MahApps.Styles.ContextMenu.VisualStudio
- Override MahApps.TextBox.ContextMenu with the VS styles

2019-10-03_18h15_14


punker76:
@mgnslndh your #Hacktoberfest changes have been merged, thanks for your contribution 👍

Issue #3376, opened 2018-11-22
NumericUpDown StringFormat doesn't handle percentages correctly.


SteffanDonal:
There's strange behaviour happening with `NumericUpDown`'s `StringFormat` property, particularly around percentages.

First it looks like the `P`, `%` & `‰` format specifiers are being ignored, but when adding those same symbols outside of a format item, the numbers appear to be divided by 100 or 1000.

I do remember more frustrating issues before however - particularly with editing values. Previously, the numbers would *look* correct initially. After editing, the real value would be multiplied repeatedly.

**To Reproduce**:

**Case 1**:
1. Set `NumericUpDown.StringFormat` to `{0:0.00%}` or `0.00%`
2. Set `NumericUpDown.Value` to `0.25`
3. See that the label now shows `0.25%` (expected `25.00%`)

**Case 2**:
1. Set `NumericUpDown.StringFormat` to `{0:0.00‰}` or `0.00‰`
2. Set `NumericUpDown.Value` to `0.25`
3. See that the label now shows `0.25‰` (expected `250.00%`)

**Case 3**:
1. Set `NumericUpDown.StringFormat` to `{0:0.0000}%`
2. Set `NumericUpDown.Value` to `0.25`
3. See that the label now shows `0.0025%` (expected `0.2500%`)
4. Additionally, editing this number causes further division to the **real value**.

**Case 4**:
1. Set `NumericUpDown.StringFormat` to `{0:0.00000}‰`
2. Set `NumericUpDown.Value` to `0.25`
3. See that the label now shows `0.00025‰` (expected `0.25000‰`)
4. Additionally, editing this number causes further division to the **real value**.

**Case 5**:
1. Set `NumericUpDown.StringFormat` to `P`
2. Set `NumericUpDown.Value` to `0.25`
3. See that the label now shows `0.25%` (expected `25.00 %`)

**Environment:**
 - MahApps.Metro version 2.0.0-alpha0083
 - OS: Win10 1803
 - Visual Studio 2017 15.7.1
 - .NET Framework 4.6.2

ckaeser-lsi:
I have an issue in the same direction. We use Numerics string format to display inline units for inputs. 

Having a string format containing a "p" or "P" results in a display of percentage
e.g StringFormat="{}{0:G3} mPa·s" input "0.986" displays as "0.00986 mPa·s"

mgnslndh:
I've investigated this issue and these are my findings:

[...]

PR #3645, opened 2019-10-11
Fix issue with P, % and ‰ used in StringFormat outside of the format item

mgnslndh:
Fix issue with P, % and ‰ used in the StringFormat property of NumericUpDown but outside of the format item. "{}{0:P}" would work but "{}{0:0.0} P" would not work correctly.

**Unit test**

Added additional asserts in appropriate unit tests.

**Related Issues**

This PR fixes a couple of problems mentioned in #3376 but does not resolve the issue completely. It only deals with using special characters outside of the format item. Other concerns in the issue is left unresolved.


punker76:
@mgnslndh your changes have been merged, thanks for your contribution 👍

Relates to #3376

PR #3664, opened 2019-10-27
Add an OverlayBrush to the SplitView

### LINKED ISSUE #3661: SplitView overriding LightDismissLayer to set opacity and color
When a SplitView is being used and the DisplayMode is set to either Overlay or CompactedOverlay whenever the pane is open or closed the storyboard sets the visibility of LightDismissLayer which functions as a overlay on the SplitView content.

By default the Rectangle is set to transparent and does not apply any Opacity.
Is there anyway to override this to have the LightDismissLayer set a different color and an Opacity so it could be used as a gray-out of the main content when the Pane opens (like Drawers in Material Design).

I can just set a Rectangle in the Content of the splitview, but it seems a waste to do that when the SplitView itself already has an overlaying Rectangle.

timunie:

- Adds a new `OverlayBrush` dependency property to the `SplitView`. 

Overlay

**Closed Issues**

Closes #3661

### DIFF
--- a/src/MahApps.Metro/Themes/SplitView.xaml
+++ b/src/MahApps.Metro/Themes/SplitView.xaml
@@ -436,7 +436,7 @@
                                             Focusable="False"
                                             IsTabStop="False" />
                             <Rectangle x:Name="LightDismissLayer"
-                                       Fill="Transparent"
+                                       Fill="{TemplateBinding OverlayBrush}"
                                        Visibility="Collapsed" />
                         </Grid>
 
--- a/src/MahApps.Metro/Controls/SplitView/SplitView.cs
+++ b/src/MahApps.Metro/Controls/SplitView/SplitView.cs
@@ -131,6 +131,23 @@
         }
 
         /// <summary>
+        /// Identifies the <see cref="OverlayBrush"/> dependency property. 
+        /// </summary>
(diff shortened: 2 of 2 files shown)


src/MahApps.Metro/Themes/SplitView.xaml
```diff
@@ -8,6 +8,7 @@
         <Setter Property="OpenPaneLength" Value="320" />
         <Setter Property="PaneBackground" Value="{DynamicResource {x:Static SystemColors.ControlLightBrushKey}}" />
         <Setter Property="PaneForeground" Value="{DynamicResource MahApps.Brushes.Text}" />
+        <Setter Property="OverlayBrush" Value="{DynamicResource MahApps.Brushes.Window.FlyoutOverlay}" />
```
punker76: @timunie I think we should leave it null, because the old Hamburger doesn’t had this.

timunie: I am ok with transparent. I think if we leave it null, it will not work anymore, or am I wrong?

timunie: @punker76 just as an inforamtion: It was `Transparent` before my changes and not `{x:null}`

punker76:

Issue #3673, opened 2019-11-04
NumericUpDown Exponent displayed for decimal type


Kabennsky:
When binding a decimal? to the value of a NumericUpDown control it still formats the value like a double with showing the exponent like 5E-05 instead of 0,00005. 

Doesn't the NumericUpDown Control just use the ToString method of the value Property?
Like in the example below a decimal ToString print the value with zeros not with exponent.
```C#
(0.00005).ToString()   // 5E-05
((decimal)(0.00005)).ToString()   // 0,00005
```

At the moment i made a workaround setting the StringFormat of the NumericUpDown controls 
```XML
<Style BasedOn="{StaticResource {x:Type mah:NumericUpDown}}" TargetType="{x:Type mah:NumericUpDown}">
    <Setter Property="StringFormat" Value="0.##########################" />
</Style>
```

But i was woundering why the NumericUpDown control behave like that.
Will the value internal be converted to a double?

timunie:
Hi @Kabennsky 
at the Moment the value is converted into a `double`, also if you bind to a `decimal`. I had a short discussion with @punker76 and he explained that we cannot change the value to `decimal` and it will cover all types, because they have different min, max and precision:
https://exceptionnotfound.net/decimal-vs-double-and-other-tips-about-number-types-in-net/ 

one idea was to add an enum where the user can choose the type of the value. The default would be `double`. At the moment this is just an idea, so will see if he gets it to work or not.

```c#
public enum ValueType
{
  Double,
  Decimal,
  Int,
  Long.
  ...
}
```

Happy coding
Tim

Kabennsky:
@timunie The Enum would work for me, would love when the control could handle different floating types. Btw I saw in the Debug output the NumericUpDown has problems with nullable types like decimal?. Nothing that is a real problem using it but I was wondering why its looked as Error:

```
System.Windows.Data Error: 6 : 'SystemConvertConverter' converter failed to convert value '<null>' (type '<null>'); fallback value will be used, if available. BindingExpression:Path=SomeMember; DataItem='SomeViewModel' (HashCode=23661558); target element is 'NumericUpDown' (Name=''); target property is 'Value' (type 'Nullable`1') InvalidCastException:'System.InvalidCastException: Das NULL-Objekt kann nicht in einen Werttyp konvertiert werden.
   bei System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
[...]

Issue #2923, opened 2017-04-13
NumericUpDown Decimal Wrong Behaviour


MurariAlex:
## What steps will reproduce this issue?
In the metro demo, texts tab, numericUpDown section, enable HasDecimals, click and keep pressed the plus button of the numeric up down under the CultureInfo dropdown until the "Stringformat, Real Value" fields gets an strange value.
Disable the HasDecimal flag, the plus button doesn't work anymore, and I don't know if the strange value is a bug itself.

![image](https://cloud.githubusercontent.com/assets/11204378/25025400/7a55cdfe-2078-11e7-8a03-6b231b67600c.png)

### Expected outcome

The NumericUpDown should keep working even after disabling the HasDecimals flag.
Maybe the value of the "Stringformat, Real Value"  is wrong.

### Environment

- MahApps.Metro __v1.5.0__
- Windows OS __10__
- Visual Studio __2015__
- .NET Framework __4.5__

punker76:
@MurariAlex This is caused by using `double?` for the Value. If you click the up or down button the Value is increased or decreased by 0.1 in this sample. A double uses floating-point arithmetic, which is approximate but more efficient, but produces e.g this `5.1 + 0.1 = 5.1999999999999993`. For exact (human) values the decimal type must be used, which is a breaking change for this control.

Add. info http://stackoverflow.com/a/1165788/920384, http://stackoverflow.com/a/618596/920384, http://net-informations.com/q/faq/float.html

/cc @xxMUROxx

PR #3678, opened 2019-11-06
Handle NumericUpDown percentage StringFormat correctly

### LINKED ISSUE #3376: NumericUpDown StringFormat doesn't handle percentages correctly.

punker76:

The current behavior for e.g. percentage text input with StringFotmat like P0 or {0:0.00%} converts the input one-to-one to the value:

100% = 100

This is not what the user expects, because the .Net default works different:

100% = 1

So this PR changes this behavior to the .Net default one.

**Unit test**

- [x] ShouldConvertTextInputWithPercentageStringFormat
- [x] ShouldConvertTextInputWithPermilleStringFormat
- [x] ShouldFormatValueInput

**Additional context**

- [x] If the stringformat is HEX the control has to accept [a..f] as Input as well.
- [x] The DemoApp crashes on an invalid (Hex) Stringformat.

mahapps_numericupdown

**Closed Issues**

Closes #3376 

/cc @mgnslndh @timunie

### DIFF
--- a/src/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/src/MahApps.Metro/Controls/NumericUpDown.cs
@@ -983,7 +983,7 @@ namespace MahApps.Metro.Controls
                 return;
             }
 
-            this.valueTextBox.Text = this.FormattedValue(newValue, this.StringFormat, this.SpecificCultureInfo);
+            this.valueTextBox.Text = this.FormattedValueString(newValue.Value, this.StringFormat, this.SpecificCultureInfo);
 
             if ((bool)this.GetValue(TextBoxHelper.IsMonitoringProperty))
             {
@@ -991,7 +991,7 @@ namespace MahApps.Metro.Controls
             }
         }
 
-        private string FormattedValue(double? newValue, string format, CultureInfo culture)
+        private string FormattedValueString(double newValue, string format, CultureInfo culture)
         {
             format = format.Replace("{}", string.Empty);
             if (!string.IsNullOrWhiteSpace(format))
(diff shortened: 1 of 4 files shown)


src/MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -983,15 +983,15 @@ private void InternalSetText(double? newValue)
                 return;
             }
 
-            this.valueTextBox.Text = this.FormattedValue(newValue, this.StringFormat, this.SpecificCultureInfo);
+            this.valueTextBox.Text = this.FormattedValueString(newValue.Value, this.StringFormat, this.SpecificCultureInfo);
```
timunie: Shouldn't we name it "PART_ValueTextBox" or something like this? I think that is a convention for WPF-Controls?

src/MahApps.Metro/Controls/NumericUpDown.cs
```diff
@@ -1006,15 +1006,36 @@ private string FormattedValue(double? newValue, string format, CultureInfo cultu
...
+
+                    // we have a format such as "N0"
+                    return ConvertStringFormatValue(newValue, format);
+                }
+            }
+
+            return newValue;
         }        
 
         private static double ConvertStringFormatValue(double value, string format)
```
mgnslndh: If we are conforming with the default .NET formatting, do we still need `ConvertStringFormatValue`?

mgnslndh: Ok, I get it :) We need to handle the conversion the other way now. 100 in text should be 1 as value.

timunie:
Hi @punker76 

some more thoughts on your PR:
- Should we derivide from `TextBox` instead of control? That way some `Properties` would be implemented by default
- If the stringformat is HEX the control has to accept [a..f] as Input as well.
- Is a `double`- value precise enough? That way you cannot use `decimal` or `ulong` - values.
- The DemoApp crashes on an invalid `Stringformat`. Is there a way to handle this? Should we handle this?

Happy coding and a 💯 times thanks. 
Tim

timunie:
Good morning @punker76 ,

> - Is a double- value precise enough? That way you cannot use decimal or ulong - values.

one addition: if we change the basetype to `decimal` we will also solve #2923 (it was closed in the past with a note that this would be a breaking change, we have a breaking change right now :smile:  ) and #3673 . 

Happy coding
Tim

src/MahApps.Metro/Controls/NumericUpDown.cs
punker76: @timunie For now I will leave it as it is. Because of further complete changes of the control itself (-> TextBox).

PR #3687, opened 2019-11-12
Resource Explorer

eiadxp:
Added resource explorer tab which will show all the resources used by Mahapps.Metro with preview of some.


eiadxp:
closed because vs 16.4 can show all resources from referenced assemblies... There is no reason for this tab to exist unless you want to explore the Mahapps.Metro resources.....

PR #3695, opened 2019-11-17
Improved external dialog to no longer extend onto other monitors

### LINKED ISSUE #3538: In full-screen mode, the dialog window pops out on 2 monitor
Good day. Desktop on 2 monitors, in full-screen mode, the dialog window pops out on 2 monitor and outside the application window

`public bool NotificationWithConfirmation(string header, string body)
        {
            var dialog = _dialogCoordinator.ShowModalMessageExternal(this, header, body, MessageDialogStyle.AffirmativeAndNegative, new MetroDialogSettings()
            {
                AffirmativeButtonText = "Yes",
                NegativeButtonText = "No"
            });
            if (dialog == MessageDialogResult.Affirmative)
            {
                return true;
            }
            else
            {
                return false;
            }
        }`
Error

**Environment(please complete the following information):**
 - MahApps.Metro version v2.0.0-alpha0391
 - OS: Win10 1607
 - Visual Studio 2019 16.1.4]
 - .NET Framework  4.7.1

AKruimink:

- Added a check for window state in `CreateModalExternalWindow`
- Added a border offset to take in account the overhang on both sides of the window when maximized causing a dialog to extend outside of the screen bounds.
- Fix `SetupExternalDialogWindow`
Use the monitor from where the owner stays and use the working area for the dialog size properties and remove also the left and right border.
- Simplify getting the working area from current monitor with new extension method `GetMonitorWorkSize`
Use this method in both variants SetupExternalDialogWindow and CreateModalExternalWindow which also fixes the main issue without calculating an offset by hand.

external-dialog-fix

**Closed Issues**

Closes #3538

### DIFF
--- a/src/MahApps.Metro/Controls/Dialogs/DialogManager.cs
+++ b/src/MahApps.Metro/Controls/Dialogs/DialogManager.cs
@@ -620,8 +620,10 @@ namespace MahApps.Metro.Controls.Dialogs
             win.Topmost = false; //It is not necessary here because the owner is setted
             win.WindowStartupLocation = WindowStartupLocation.CenterOwner; //WindowStartupLocation should be CenterOwner
 
+            var offset = SystemParameters.WindowNonClientFrameThickness.Left + SystemParameters.WindowResizeBorderThickness.Left +
+                    SystemParameters.WindowNonClientFrameThickness.Right + SystemParameters.WindowResizeBorderThickness.Right;
             //Set Width and Height maximum according Owner
-            win.Width = window.ActualWidth;
+            win.Width = window.WindowState != WindowState.Maximized ? window.ActualWidth : window.ActualWidth - offset;
             win.MaxHeight = window.ActualHeight;
             win.SizeToContent = SizeToContent.Height;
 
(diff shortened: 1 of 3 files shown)


AKruimink:
Few things.

1. With the way i currently do it i calculates the offset always, this could still be changed to only get calulated on Maximized state, any preference?

2. `SetupExternalDialogWindow` seems to maximize at all times, i wasnt sure if it required the change as well and thus has been left out, a minor issue with the current setup is when a user has monitors with different resolutions the `PrimaryScreenWidth` might not fit.

AKruimink:
Changed the check regardless, might be less pretty to look at, but it's not worth creating an offset each time when most of the time it's not used anyways.

AKruimink:
@punker76 or @timunie   The `CreateModalExternalWindow` has been fixed and seems to be working.

For the `SetupExternalDialogWindow` i can't really find a proper way of doing it. The only ways i could come up with us by using `Screen` but would require `Windows.Forms` which i think is a bad idea, or you end up having to wrap around pinvoke and do it that way, which doesnt seem great either.

So the only options im further left with is.
1. leave it as is, `SystemParameters.PrimaryScreen(Width/Height)` will be set, and if a user has a smaller secondary screen you will have the overlap onto the other monitor (or the oppositeif the screen is bigger).

2.  Add optional parameters to the method that allows for custom sizes, or if left empty just use the `SystemParameters.PrimaryScreen(Width/Height)` .

option 2 would not persé fix the issue but would offer more flexibility at the least.

Any input would be appreciated

punker76:
@AKruimink I have a solution for it and will check in later.

punker76:
@AKruimink your changes have been merged, thanks for your contribution 👍

PR #3696, opened 2019-11-18
DatePicker enhancements

### LINKED ISSUE #3656: Clear Button in DatePicker
Hi.. it will be nice to have a clear button in Date Picker. 

Some times we want to select the date and proceed 
Some time we don`t want to select the date and proceed.

We can edit the text when it is read only is set to false for the picker but it is not possible when read only is true in picker

punker76:

- Add `DatePickerHelper` with new attached properties. This allows us to handle this button separately to another clear button.
  - `DropDownButtonContent`
  - `DropDownButtonContentTemplate`
  - `DropDownButtonFontFamily`
  - `DropDownButtonFontSize`
- Use new DatePicker helper for `DateTimePicker` and `TimePicker`
- Introduce `MahAppsCommands.ClearControlCommand`
  - This command does the same as the attached TextBoxHelper IsClearTextButtonBehaviorEnabled property and should be used later for all controls too. So then we doesn't need 2 separate styles for ButtonCommand usage.
- [x] Implement clear button for `DatePicker`
- [x] Implement clear button for `DateTimePicker` and `TimePicker`

image

**Unit test**

-/-

**Closed Issues**

Closes #3656

### DIFF
--- a/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/DateExamples.xaml
+++ b/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/DateExamples.xaml
@@ -34,18 +34,24 @@
                 <DatePicker Width="200"
                             Margin="{StaticResource ControlMargin}"
                             HorizontalContentAlignment="Stretch"
+                            Controls:TextBoxHelper.ClearTextButton="True"
                             Controls:TextBoxHelper.UseFloatingWatermark="True"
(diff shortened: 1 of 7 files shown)


timunie:
Hi @punker76 ,
Thinking further I think this Helper class could be used for `ComboBox` `DropDownButton` etc in the future. So maybe we should call it more general like `DropDownHelper`.

Happy coding
Tim

PR #3704, opened 2019-11-28
Allow using the DropDownButton without a dropdown menu

floele-sp:
If it is necessary to use the DropDownButton without a menu under some circumstances (which is reasonable, because you can also hide the arrow button already) and only the "Click" event is used, the button will still try to open the empty menu, which can be seen by the user as a thick black line.

This change only tries to open the menu if there actually are some items inside and makes the DropDownButton behave similar to a normal button when clicked.


punker76:
@floele-sp your changes have been merged, thanks for your contribution 👍

PR #3723, opened 2020-01-04
Changing blue color scheme to blue

batzen:
This PR changes to blue color scheme to a "real" blue.
@punker76 we already chatted about this ;-)

### DIFF
--- a/src/MahApps.Metro/Styles/Themes/GeneratorParameters.json
+++ b/src/MahApps.Metro/Styles/Themes/GeneratorParameters.json
@@ -110,11 +110,11 @@
     {
       "Name": "Blue",
       "Values": {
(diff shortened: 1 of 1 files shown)


timunie:
Hi @batzen ,
I liked the old blue value. I'm with you that blue should be blue but may we keep the old `Theme` under a new Name? What do you think? 

P. S. : If I'm the only one I don't mind this change because I generate my `Themes` at runtime. I just thought others might want to keep their Theme as well. 

Happy coding 
Tim

batzen:
Adding new color variants is very easy.
But the current/old blue is in fact cyan.
image
Left: current/old blue
Right: cyan

timunie:
@batzen thank you this ia a really good point. :+1:

punker76:

PR #3771, opened 2020-04-06
Switching to ThemeManager from ControlzEx

batzen:
This PR removes the ThemeManager and switches everything to the general purpose ThemeManager from ControlzEx.
This allows seamless sync of themes across different libraries, for example MahApps.Metro and Fluent.Ribbon.
This "new" ThemeManager also contains the ability to generate themes at runtime while taking the current windows settings into account and allowing users to choose to use HSL colors for accents etc.. HSL colors don't use the alpha channel for the accent color steps, but tinted colors instead.


batzen:
@timunie FYI

punker76:
@batzen thx mate

tokm:
I can't seem to find any documentation for the transition from ThemeManager to ControlzEx.  I used MahApps.Metro.ThemeManager.AddAccent in my project.  How do I do the same thing now since ThemeManager in MahApps.Metro has been removed?

timunie:
@tokm see: https://github.com/MahApps/mahapps.com/issues/8

PR #3775, opened 2020-04-16
(GH-3746) Fix Date and Time selection and other improvements for DateTimePicker

### LINKED ISSUE #3746: DateTimePicker changes date to 1/1/0001 when selecting time first
Firstly great work!

Using DateTimePicker in 2.0.0-alpha0660

Selecting a time from the dropdown before selecting the date changes the date to 01/01/0001.

DateTimePicker is bound to a nullable date time.  Same occurs when DateTimePicker is not bound to anything at all actually.

It works fine in 1.6.5 where selecting the time first still results in the textbox saying "select a date"

I must target .net core, so sadly can't use 1.6.5

Here is an example of both

[Doesnt work using 2.0.0 alpha](https://github.com/didster/DateTimePickerDoesntWork)
[Works using 1.6.5](https://github.com/didster/DateTimePickerWorks)

timunie:

- Fix input for date and time text input for DateTimePicker
  + if the Time is selected before a Date is selected use DateTime.Today instead of DateTime.Min
  + If a new Value is selected from the Calendar we use the already Selected time.
- Use PopupEx to fix position change for floating watermark
- Fix wrong focus for popup content
- Fix closing (unfocusing) and opening PopUp
- Fix capturing CalendarItem
- Add key processing if Popup is open
- Add DisplayDateChanged event handling
- Respect IsEnabledProperty
- Remove TimePickerBaseSelectionChangedEventArgs and use RoutedPropertyChangedEventHandler and RoutedPropertyChangedEventArgs
- Set selected date and time from textbox

**Unit test**

DateTimePickerTests

**Additional context**

~~There is one issue left where I have no idea how to solve this: If the user inserts a new `DateTime` manually where the time is `00:00:00` the time might get overridden with the last used time. I think this case is rather rare, but should be considered anyway.~~

**Closed Issues**

Closes #3746


timunie:
@punker76 can we change `TimePickerBaseSelectionChangedEventArgs` to `GenericSelectionChangedEventArgs`? We then could reuse it e.g. for the `ColorPicker`. It is already generic, just the name does not say it :-)

Have a great day
Tim

punker76:
> @punker76 can we change `TimePickerBaseSelectionChangedEventArgs` to `GenericSelectionChangedEventArgs`? We then could reuse it e.g. for the `ColorPicker`. It is already generic, just the name does not say it :-)
> 
> Have a great day
> Tim

@timunie You can use the RoutedPropertyChangedEventHandler and RoutedPropertyChangedEventArgs which are generic classes. I removed the TimePickerBaseSelectionChangedEventArgs because we don't need them.

timunie:
Thank you for the info. I wil use this instead

PR #3781, opened 2020-04-26
Removing test that already exist in ControlzEx

batzen:
This PR removes all test for `ThemeManager` that already exist in ControlzEx.
That way updating ControlzEx should be easier in the future.


punker76:

PR #3782, opened 2020-04-26
Add IsVisible property to HambugerMenuItemBase

### LINKED ISSUE #3768: Toggling visibility on hamburger menu
It would be nice to have a visibility open on hamburger menu items to hide from people that are not suppose to see instead of just it being disabled for certain users.

bNobo:
I've added a Visibility property to HamburgerMenuItem and HamburgerMenuSeparatorItem in response to issue #3768 

**Closed Issues**
It will closes #3768

### DIFF
--- a/src/MahApps.Metro/Controls/HamburgerMenu/MenuItems/HamburgerMenuItemBase.cs
+++ b/src/MahApps.Metro/Controls/HamburgerMenu/MenuItems/HamburgerMenuItemBase.cs
@@ -10,9 +10,9 @@ namespace MahApps.Metro.Controls
         public static readonly DependencyProperty TagProperty = DependencyProperty.Register(nameof(Tag), typeof(object), typeof(HamburgerMenuItemBase), new PropertyMetadata(null));
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


timunie:
Hi @bNobo ,

is there a reason you did not implement this in the base class? I am sorry if you got me wrong, I wanted to confirm your first idea.

bNobo:
Hi @timunie ,

you're absolutely right, I'll move the property to the base class to avoid duplication. I've also realized I forgot to update the template of other `HamburgerMenuItemBase` descendants like `HamburgerMenuHeaderItem`. I will fix that too.

bNobo:
Everything should be OK now, Visibility property works for every `HamburgerMenuItemBase` descendants. Feel free to make me know if I still can improve the code in any way.

punker76:
@bNobo your changes have been merged, thanks for your contribution 👍

I've changed the property from Visibility to IsVisible. IMO IsVisible is the better way to say if an item should be shown or not. With the Visibility property you can set an item also to hide which means that the item is only hidden.

PR #3804, opened 2020-05-21
Fixed Validation Tooltip that never shows up.

luisfb:
I was having an issue where the canShow variable always ends up false.
This change should fix the problem and not interfere in the current logical flow.


punker76:
@luisfb Do you have an example for this?

luisfb:
Yes. I recorded a video.
First take a look at my ViewModelBase and also de code that adds the error:
ViewModel Base
Error Command

Here is the video of the NOT Working pop-up error:

Not Working video

After my change (this pull request), here is the result:

Working video

punker76:
@luisfb your changes have been merged, thanks for your contribution 👍 Thx for using the latest alpha hot stuff and reporting the issue.

PR #3815, opened 2020-05-26
Fixed #3811: Effect of parent's IsEnabled on DropDownButton

### LINKED ISSUE #3811: Regression in v2.0: Effect of parent's IsEnabled on DropDownButton
**Description of the bug**
If a parent containing a `DropDownButton` has a bound `IsEnabled` property, and the property is switched to disabled and enabled again, the `DropDownButton` remains disabled. This only happens in v2.0.0 but works as expected in v1.6.5.

**Expected behavior**
The `DropDownButton` should be enabled again when parent's `IsEnabled` is set to `true`.

**Environment:**
 - MahApps.Metro version v2.0.0
 - Win10 1909
 - Visual Studio 2019 16.6.0
 - .NET Framework 4.7.2

**Repo**
https://github.com/yjagota/MahAppsTest

yjagota:
**The changes made in this pull request**

This pull request fixes a regression in v2.0 where changing a parent's IsEnabled to true does not enable the DropDownButton.

Closed Issues
#3811

### DIFF
--- a/src/MahApps.Metro/Themes/DropDownButton.xaml
+++ b/src/MahApps.Metro/Themes/DropDownButton.xaml
@@ -34,6 +34,7 @@
         <Setter Property="MinHeight" Value="26" />
         <Setter Property="Padding" Value="3" />
         <Setter Property="SnapsToDevicePixels" Value="True" />
+        <Setter Property="IsEnabled" Value="{Binding IsEnabled, RelativeSource={RelativeSource FindAncestor, AncestorType={x:Type FrameworkElement}}}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="{x:Type Controls:DropDownButton}">


punker76:
@yjagota Thanks for trying to solve the issue. Binding to the IsEnabled property of the parent is not a good approach and doesn't solve the real bug. The issue comes from the inner Button which set's the IsEnabled property twice. After that the control IsEnabled property will never get an update. To fix this I implemented the ICommandSource for DropDownButton and SplitButton in #3819 which fixes the issue.

yjagota:
Yes. I saw the commit. Sorry for the half baked fixes for both issues. These both were my first pull request submissions, so I learned how to submit pull requests and also learned that I should be way more careful when posting them. :) And thanks again for fixing the issues and giving us this awesome package. I hope I can make a worthwhile contribution to the project someday. Take care and stay safe.

Issue #3818, opened 2020-05-26
Idea collection for a new demo App


timunie:
**Is your feature request related to a problem? Please describe.**
Now that the v2.0.0 is out maybe the Demo App should be updated as well. I would like to open this feature request to collect some ideas / drafts

**Describe the solution you'd like**
These are my ideas so far: 
- Use the `HamburgerMenu` instead of a `TabControl`. 
- Use the same Layout for almost every Control
- Provide a Link to the online doc where possible (see: #3053 )
- Provide sample `XAML` to copy & paste it into your own app 

**Additional context**
I have implemented this into my personal Controls-Libary and here is an early preview of it: 
![DemoAppExample](https://user-images.githubusercontent.com/47110241/82902539-e15e8900-9f5f-11ea-9382-f1bb38862eff.png)

To create this view:
```xaml
<local:ExampleView_Base x:Name="demoView" xml:space="preserve">
    <local:ExampleView_Base.ExampleXaml><![CDATA[<ctrls:CircularProgressBar 
        Value="[Value]" 
        Minimum="[Minimum]"
        Maximum="[Maximum]"
        Width="[Width]"
        Height="[Height]"
        HorizontalAlignment="[HorizontalAlignment]"
        VerticalAlignment="[VerticalAlignment]"
        Content="[Content]"
        ContentStringFormat="[ContentStringFormat]" />
        ]]></local:ExampleView_Base.ExampleXaml>
    <ctrls:CircularProgressBar x:Name="MyProgressBar"
                                Width="200"
                                Height="200"
                                HorizontalAlignment="Center"
                                VerticalAlignment="Center"
                                ContentStringFormat="0' %'"
                                Maximum="100"
                                Minimum="0"
                                Value="5" />
</local:ExampleView_Base>
```

and in the Constructor:
```c#
public RadialProgressExample()
{
    InitializeComponent();

    demoView.AddDemoProperty(CircularProgressBar.ValueProperty, MyProgressBar);
    demoView.AddDemoProperty(CircularProgressBar.MinimumProperty, MyProgressBar);
    demoView.AddDemoProperty(CircularProgressBar.MaximumProperty, MyProgressBar);
    demoView.AddDemoProperty(CircularProgressBar.IsFilledProperty, MyProgressBar);
    demoView.AddDemoProperty(CircularProgressBar.IsIndeterminateProperty, MyProgressBar);

[...]

PR #3825, opened 2020-05-28
[WIP] Proposal of an extended Demo App

timunie:
The idea for the demo recieved more positive votes than I thought. For this reason I implemented what I have so far here. 

Note: It is still a first draft for discussion

**Unit test**
None

**Additional context**
Any discussion is very welcome

**Closed Issues**
#3818

### DIFF
--- a/src/MahApps.Metro.Demo_v2/App.xaml
+++ b/src/MahApps.Metro.Demo_v2/App.xaml
@@ -1,0 +1,34 @@
+﻿<Application x:Class="MahApps.Metro.Demo_v2.App"
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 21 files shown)


punker76:
Tim, this has been sitting here since 2020, and the funny part is how much of it we are about to use anyway.

I am closing it, and not because it was wrong. The demo has moved on since, so getting this branch onto today's `develop` would be more work than carrying the ideas across, and carrying them across is the plan: the demo view with the control on one side and its properties on the other, an editor per property type, the link to the online docs, a navigation with groups instead of one tab per control. #3818 stays open as the place for all of that, and what I would like to build there is a demo app of its own shape, along the lines of the WinUI 3 Gallery.

One more thing. Back then @AKruimink pointed at ShowMeTheXAML and you said it was on the list. It is off the list now: #4648 puts it into the demo, and the XAML that is shown is the XAML that runs, with no second copy to keep in step. Which was your point about copy and paste, six years early.

Thanks for the push, it did land. Grüße an die Avalonia-Fraktion. 😉

PR #3873, opened 2020-07-07
Fix ThemeManager changes not working with different (STA) threads

### LINKED ISSUE #3872: Theme Manager wont work with 2 Main windows on different (STA) Threads
I have a Prism Application with a Bootstrapper and App.MainWindow (MetroWindow) running on MainThread and a seperate Login Window (also MetroWindow) running on sperate STA Thread to keep the UI responsive while Bootstrapper is running.

When switching the Theme in any situation the Application throws an exception: "The calling thread cannot access this object because a different thread owns it."

image

image

My assumption is that static Implementation and Event subscription from MetroWindow causes the Metro Window to crash when it does not belong to the Mainthread.

Thanks in advance,
Johannes

punker76:

Execute ThemeManager changes at the current Dispatcher UI thread.

## Closed Issues

Closes #3872

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -6,7 +6,7 @@
 #tool "dotnet:?package=NuGetKeyVaultSignTool&version=1.2.18"
 #tool "dotnet:?package=AzureSignTool&version=2.0.17"
 
-#tool GitVersion.CommandLine&version=5.0.1
+#tool GitVersion.CommandLine&version=5.3.7
 #tool gitreleasemanager
 #tool xunit.runner.console
 #tool vswhere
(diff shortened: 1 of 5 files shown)


amkuchta:
@punker76 is there any way that we can get this into the next release instead of 2.2.0? Not sure if you are using Semantic Versioning or not, but this seems like a bugfix to me, not a breaking feature that requires a minor version update.

Also, I really want it 😂

timunie:
@amkuchta the PR is already merged so I think it will be in the next release. Maybe there will be no 2.1.2 because of another change which adds a new feature. 

Happy coding
Tim

PR #3889, opened 2020-07-23
Fixes for MathConverter

### LINKED ISSUE #3888: MathConverter should either implement convert back or return Binding.UnsetValue
## Describe the problem
I binded a `Slider` to a value which I converted using the `MathMultiplyConverter`. it was shown correctly but never reported the changed value back to the source. It even did not throw any `Exception` so I didn't knew it would not work until I had a look into the source of MahApps

## Soulution option 1
All `MathConvter` should implement a `ConvertBack`-Method. Example: `value+parameter=result` will be converted back to `result-parameter=value`.

## Solution option 2
All `MathConvter` should return `Binding.UnsetValue` indestead of `Binding.DoNothing`. So if the user tries to implement a twoway bindig he will get notified that it is not supported. 

Happy coding
Tim

punker76:

- Check for 0 on division and trace it as warning
- Return DependencyProperty.UnsetValue at convert back methods

## Unit test

- MathConverterTests

## Additional context

Maybe the ConvertBack can be implemented, but not in this PR.

## Closed Issues

Closes #3888

### DIFF
--- a/src/MahApps.Metro/Converters/MathConverter.cs
+++ b/src/MahApps.Metro/Converters/MathConverter.cs
@@ -110,7 +110,7 @@ namespace MahApps.Metro.Converters
     [MarkupExtensionReturnType(typeof(MathAddConverter))]
     public sealed class MathAddConverter : MarkupMultiConverter
     {
-        private static readonly MathConverter MathConverter = new MathConverter() { Operation = MathOperation.Add };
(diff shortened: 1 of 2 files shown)


timunie:
@punker76 that was lightning fast :zap: . Maybe I should have said that I have already a working solution in my case. 

Thank you very much!

PR #3973, opened 2020-11-21
Data grid column style helper extension memory leak

### LINKED ISSUE #3972: Memory leak created by the ColumnStylesHelper
## Describe the bug

The [DataGridColumnStylesHelperExtension](https://github.com/MahApps/MahApps.Metro/blob/c36f14da09ff70630d9e558a6f4564c34c176e9e/src/MahApps.Metro/Controls/Helper/DataGridColumnStylesHelperExtension.cs#L31) is [attached](https://github.com/MahApps/MahApps.Metro/blob/9f7344170ba5ed7c04775b597c6edcb4efb7bfcc/src/MahApps.Metro/Styles/Controls.DataGrid.xaml#L771) but never [detached](https://github.com/MahApps/MahApps.Metro/blob/c36f14da09ff70630d9e558a6f4564c34c176e9e/src/MahApps.Metro/Controls/Helper/DataGridHelper.cs#L627) from the grid, leaving a memory leak with the [Columns.CollectionChanged](https://github.com/MahApps/MahApps.Metro/blob/c36f14da09ff70630d9e558a6f4564c34c176e9e/src/MahApps.Metro/Controls/Helper/DataGridColumnStylesHelperExtension.cs#L41) handler. This causes the whole window that is hosting the data grid to remain in memory. 

## Steps to reproduce

1. [Replace](https://github.com/lipchev/MahApps.Metro/commit/47ae93fe438cf94019eefd94de42c669f28fd837) the albums list view in the CleanWindowDemo.xaml with an equivalent data grid  
2. Open and close the CleanWindowDemo

## Expected behavior

When the parent is unloaded the memory is liberated

## Actual behavior

The CleanWindowDemo window is retained in memory via the grid and the DataGridColumnStylesHelperExtension (see screenshot below)
Repeating the same steps with the `mah:DataGridHelper.ColumnStylesHelper="{x:Null}"` and the problem no longer occurs

## Environment

```
MahApps.Metro version: v2.0.0
Windows build number: Win10 1909 [Version 10.0.18363.836]
Visual Studio: 2019 16.6.0
Target Framework: .Net Core 3.1 (v4.5.2, v4.7.2)
```

## Screenshots

image

lipchev:

- [x]  CleanWindowDemo: replaced the list view with a data grid in order to demonstrate the memory leak (I can revert this if you want)

- [x] Added more error checking for the Attach/Detach methods (e.g. calling detach before attach or attaching the same instance to another grid)
- [x] Returning a new instance of the DataGridColumnStylesHelperExtension in the ProvideValue method and left a remark about the issue with returning 'this' 

## Unit test

- [ ] Tests of the Attach/Detach methods could be devised (if necessary)

## Additional context

I've used the dotMemory profiler before and after- the CleanWindowDemo is now properly freed from memory (even if the Detach() method is never called)

## Closed Issues

Closes #3972


punker76:
@lipchev your changes have been merged, thanks for your contribution 👍

PR #4021, opened 2021-01-08
Implement ColorToSolidColorBrushConverter

### LINKED ISSUE #4019: ColorCanvas has 14 Binding Expression Errors
## Describe the bug
Using the MahApps `ColorCanvas` results in 14 binding expressions errors when the control is displayed. VS2019 shows them in the new XAML Binding Failures tool (Debug > Windows > XAML Binding Failures). Alternatively, you add a `TraceListener` via `PresentationTraceSources.DataBindingSource` (this is what we use during development to avoid binding errors).

## Steps to reproduce

Create a new WPF application (I used framework, but probably doesn't matter) and use a `ColorCanvas`, e.g. using the sample code from the [MahApps ColorPicker Documentation](https://mahapps.com/docs/controls/ColorPicker#colorcanvas).

```xml
<Window x:Class="MahAppsDemo.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:MahAppsDemo"
        xmlns:mah="http://metro.mahapps.com/winfx/xaml/controls"
        mc:Ignorable="d"
        Title="MainWindow" Height="450" Width="800">
    <Grid>
        <mah:ColorCanvas x:Name="ColorCanvasExample"
                         SelectedColor="Blue"
                         DefaultColor="Transparent"
                         LabelAlphaChannel="Alpha"
                         LabelBlueChannel="Blue"
                         LabelGreenChannel="Green"
                         LabelRedChannel="Red" />
    </Grid>
</Window>
```

## Expected behavior
No binding expression errors occur.

## Actual behavior
14 binding expressions errors show up (see screenshot below). Using a `TraceListener` in combination with `Debugger.Break();` calls completely kills the development experience.

## Environment

```
MahApps.Metro version: v2.4.3
Windows build number: Win10 2004
Visual Studio: 2019 16.8.3
Target Framework: v4.7.2
```

## Screenshots

image

timunie:
Implement a `ColorToSolidColorBrushConverter` because if we bind direct to a `Color` we might get binding issues.

## Unit test
none

## Additional context
This is an issue for some users but not for all. 

@nzain can you clone my branch, attach your tracing-tool and give me a feedback if it works now for you as well? Thank you. 

@seba30 if you enter a wrong color name I want to raise an error so the user gets notified that the value was wrong. So in this case the binding issue is wanted. 

## Closed Issues
Closes #4019

### DIFF
--- a/src/MahApps.Metro/Converters/ColorToSolidColorBrushConverter.cs
+++ b/src/MahApps.Metro/Converters/ColorToSolidColorBrushConverter.cs
@@ -1,33 +1,37 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Data;
 using System.Windows.Media;
+using JetBrains.Annotations;
 
 namespace MahApps.Metro.Converters
 {
     /// <summary>
     /// Converts a given <see cref="Color"/> into a <see cref="SolidColorBrush"/>.
     /// </summary>
-    [ValueConversion (typeof(Color), typeof(SolidColorBrush)) ]
-    public class ColorToSolidColorBrushConverter : MarkupConverter
+    [ValueConversion(typeof(Color), typeof(SolidColorBrush))]
+    public class ColorToSolidColorBrushConverter : IValueConverter
     {
-        static ColorToSolidColorBrushConverter _DefaultInstance;
+        private static ColorToSolidColorBrushConverter defaultInstance;
 
         /// <summary>
-        /// returns a static instance if needed.
+        /// Gets a static instance of the converter if needed.
         /// </summary>
-        public static ColorToSolidColorBrushConverter DefaultInstance => _DefaultInstance ??= new ColorToSolidColorBrushConverter();
+        public static ColorToSolidColorBrushConverter DefaultInstance => defaultInstance ??= new ColorToSolidColorBrushConverter();
 
         /// <summary>
-        /// Gets or Sets the FallbackBrush which should be used if the conversion fails. 
+        /// Gets or Sets the brush which will be used if the conversion fails.
         /// </summary>
+        [CanBeNull]
         public SolidColorBrush FallbackBrush { get; set; }
 
-        protected override object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        /// <summary>
+        /// Gets or Sets the color which will be used if the conversion fails.
+        /// </summary>
+        [CanBeNull]
+        public Color? FallbackColor { get; set; }
+
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
         {
             if (value is Color color)
             {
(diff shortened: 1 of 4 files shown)


nzain:
Unfortunately, it doesn't even build on my machine. VS complains about net4.7 SDK missing (we use 4.7.2 for years) and VS doesn't show the csproj file. Neither can I change the SDK (combobox empty and disabled). I can't test without significant time investment and, to be honest, I've dropped the idea of using MahApps after some more obstacles.

If you want to reproduce it: A simple trace listener looks like this

```cs
    public sealed class BreakDebuggerTraceListener : TraceListener
    {
        public override void Write(string message)
        {
        }

        public override void WriteLine(string message)
        {
            Debugger.Break();
        }
    }
```
and this is how you enable it for your application before startup:

```cs
    TraceSource dataBindingSource = PresentationTraceSources.DataBindingSource;
    dataBindingSource.Listeners.Add(new BreakDebuggerTraceListener());
```

punker76:
> I've dropped the idea of using MahApps after some more obstacles

@nzain Can you say what other obstacles are you mean?

nzain:
@punker76 We have a large WPF desktop (framework) app and we have our own styles. Not really using themes, just basic WPF with a few changes to the defaults. We used a small fraction of Xceed's extended wpf toolkit comunity edition (one SplitButton, one ColorCanvas, a few NumericUpDown, many many BusyIndicator occurences). We always had problems with Xceed's toolkit and upgrading this nuget package always broke the AvalonDock for many versions and for different reasons every time. Thus we were stuck on a quite old version. As you may know, they've changed their license and that finally forces us to find alternatives.

We've successfully migrated to Dirkster's AvalonDock fork (works much better).
We came up with our own BusyIndicator implementation (a core component of our UI concept with many long running tasks).

I thought MahApps could fill the remaining gap (ColorPicker, SplitButton, NumericUpDown), but MahApps heavily revolves around its own style (and it doesn't match ours). Try to use the `ColorPicker` without MahApps styles - buttons are empty, etc. Personally, I find the `NumericUpDown` button order confusing (increment left, decrement right). Additionally, we can't allow editing of the alpha channel (our customers would confuse themselves) but the MahApps ColorPicker is rather new and not configurable in this aspect. I could have created a PR for the last one, but that wouldn't solve all my issues right? Don't take it too serious - I guess MahApps is great when you use it right from the start. However, it can't fill a certain gap in a large existing application.

timunie:
Hi @nzain 

Thank you for providing this simple class. The debugger breaks if I enter an invalid color name but IMO this should be the case. All other binding issues seems to be resolved now. 

Let me say a few words to your other issues:
- I think you can use MahApps without using all styles, but you will need to import the used controls on your own. Moreover you may want to adopt the `ControlTemplate` to suit your design. MahApps is open source so you can copy the original `Styles` 
- `NumericUpDown` -> Just use `SwitchUpDownButtons=true`
- I am working on hiding the `Alpha-Channel` as it seems to be a common need: see #4022 

> Don't take it too serious - I guess MahApps is great when you use it right from the start. However, it can't fill a certain gap in a large existing application.

Thank you for clarifying this more as it sounded like "MahApps is total mess". :+1: 

Happy coding 
Tim

nzain:
Sorry for that! I'll keep an eye on MahApps 👍

punker76:

PR #4090, opened 2021-04-23
Implement `DefaultValue` to the `NumericUpDown`-Control

### LINKED ISSUE #3668: NumericUpDown must set minimum on value deleting
**Describe the bug**
If the minimum value is set, then when deleting the contents (Delete or Backspace button), the minimum value must be set. Instead, a conversion error is displayed.
The problem is observed when using mvvm

**To Reproduce**
See demo-project

**Screenshots**
image

**Environment(please complete the following information):**
 - MahApps.Metro version - v1.6.5
 - OS: Win10
 - .NET Framework - 4.7

**Repo**
https://github.com/Pekshev/MahAppsTestNumericUpDown

### LINKED ISSUE #3936: NumericUpDown shouldn't be able to be left empty
## Describe the feature

When you delete the value in a NumericUpDown control a message will appear telling the user the value '' couldn't be converted. That's fine, but in my opinion the control should go back to its previous value as soon as its losing focus and if the textbox is still empty. It already does that when you replace a number by a space, but not when the string is completely empty.
Alternatively it could go back to a user defined default value.

timunie:
Implemented a new Property `DefaultValue` which will be used if the `Value` is `null`. This will also prevent errors for non-nullable values if set.

## Unit test
Implemented some basic tests which worked fine so far. 
Test-Overview: 
```c# 
MahApps.Metro.Tests.NumericUpDownTestsNumericUpDownTests.ShouldSetDefaultValue
// 1. Test: The default value must be set here. Let's check this.
// 2. Test: There is no default value, so we should be able to set it to null
// 3. Test: We set the Default Value greater than the Maximum. It should be corrected by the control
// 4. Test: We set the Default Value lower than the Minimum. It should be corrected by the control
```

## Additional context
See the demo for Testing

image

## Closed Issues

Closes #3668 
Closes #3936

### DIFF
--- a/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/TextExamples.xaml
+++ b/src/MahApps.Metro.Samples/MahApps.Metro.Demo/ExampleViews/TextExamples.xaml
@@ -64,7 +64,7 @@
 
         <GroupBox Grid.Row="0"
                   Grid.Column="0"
-                  Margin="4 2"
+                  Margin="4,2"
                   Header="TextBox">
             <AdornerDecorator>
                 <StackPanel>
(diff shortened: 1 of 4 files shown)


punker76:

francoscarpa:
Is this feature ready to be used?

timunie:
I think so. If not, it will be available in the next release.

Happy coding
Tim

francoscarpa:
I updated my project to use the latest version of the MahApps.Metro Nuget package, but still I can't use it.

punker76:
@francoscarpa It will be available in v3.0

Lucaa83:
> @francoscarpa It will be available in v3.0

When will v3.0 be released?

Issue #4096, opened 2021-04-29
Flyout color when change color in a single MetroWindow


recetarioMDWiki:
## Describe the bug

Hello everyone!
I have a problem with changing the color to a single window. I will try to explain it with an example.

## Expected behavior

I was hoping to change the color of the secondary window without affecting the main one

## Actual behavior

In my Application.xaml, in ResourcesDictionary, I have this:
```
<ResourceDictionary.MergedDictionaries>
    <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Controls.xaml" />
    <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Fonts.xaml" />
    <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Themes/Light.Blue.xaml" />
    <ResourceDictionary Source="/Resources/Icons.xaml" />
    <ResourceDictionary Source="pack://application:,,,/MahApps.Metro;component/Styles/Controls.TabControl.xaml" />
</ResourceDictionary.MergedDictionaries>
```

The user can choose which color he wants to apply. In the example, the user has chosen Pink.
Starting the application, I apply the theme and color that the user has saved in the configuration, using this:

```ThemeManager.Current.ChangeTheme(Application.Current, My.Settings.EstiloHUM, My.Settings.ColorHUM)```

![image](https://user-images.githubusercontent.com/50943477/116546245-aee8a900-a8f1-11eb-9f90-9263fa8b066e.png)

When I open the secondary window, according to a series of characteristics, that window may want to open it in Steel color.
I use this:

```ThemeManager.Current.ChangeTheme(Me, "Light.Steel")```

So far so good.

![image](https://user-images.githubusercontent.com/50943477/116546315-c1fb7900-a8f1-11eb-8b97-74e5dff24d71.png)

But if I look in the main window, everything is still Pink except for this flyout (which has Theme = "Accent" in the .xaml). This and other flyouts from main window are in design code (.xaml) and I used them as filters for different sections.

![image](https://user-images.githubusercontent.com/50943477/116546371-cfb0fe80-a8f1-11eb-9571-80f34a9082f2.png)

And also, when I close the second window, the main one seems to go back to Pink, but that flyout continues to show with Steel, and others flyouts, which has Theme = "Adapt", the content shows like this:

[...]

PR #4112, opened 2021-05-24
Fix for #4096

batzen:

We have to use the correct theme when we receive a ThemeChanged event as it's a global event.
If the event targets the current window we use the theme from e.NewTheme, otherwise we have to detect the theme for `this` and use that as the new flyout theme.

## Closed Issues

#4096


punker76 (approved):
LGTM 👍🏻

punker76:

PR #4147, opened 2021-07-01
Fix compile

batzen:
Caused by changed invocation order with SDK version 5.0.301...

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.csproj
+++ b/src/MahApps.Metro/MahApps.Metro.csproj
@@ -46,20 +46,20 @@
         </EmbeddedResource>
     </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


batzen:
@punker76 The changes you made break building in Rider as `IsCrossTargetingBuild` is always false in Rider.

FroggieFrog:
build via `dotnet build`: ✔️ 
build via `VS 2019`: ✔️
build via `VS 2022`: ✔️
build via `Rider 2021.1.3`: ✔️

tests succesful: ❌ 

test_fail_NumericUpDownTests

punker76:
@batzen Rider should now work again
@FroggieFrog Yeah, I saw these failed tests too (only on .NET 5 target). I don't know why, because the appveyor build is working...

FroggieFrog:
After the last commit everything is still compiling.

I dug a little because of those failing tests and it seems the framework changed the `NumberFormat` or the culture changed it.
Testdata is using `/` and culture in .NET 5 is using `,`.

test_fail_wrong_format

PR #4187, opened 2021-10-02
Adding proper version range for ControlzEx dependency

batzen:
We have to constrain the version range for our dependency on ControlzEx. Otherwise people might end in a situation where nuget automatically chooses an incompatible higher version.

### DIFF
--- a/src/MahApps.Metro/MahApps.Metro.csproj
+++ b/src/MahApps.Metro/MahApps.Metro.csproj
@@ -16,7 +16,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="ControlzEx" version="5.*" />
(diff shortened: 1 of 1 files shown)


dittodhole:
just chime in to say that this bugfix is not reflected in [v2.4.8](https://www.nuget.org/packages/MahApps.Metro/2.4.8):

> ControlzEx (>= 4.4.0 && < 5.0.0)

dittodhole:
@punker76 ah, I see 👍  Would it be an option, to target netcore3.0 with [ControlzEx v4](https://www.nuget.org/packages/ControlzEx/4.4.0) and netcore3.1 with [ControlzEx v5](https://www.nuget.org/packages/ControlzEx/5.0.1)?

Issue #4217, opened 2021-12-01
Binding Information messages - slowing down performance (System.Windows.Data Information: 10)


oriches:
I am seeing alot of Information level binding messages when I turn on the Debug Listener in Snoop:

_System.Windows.Data Information: 10 :
Cannot retrieve value using the binding and no valid fallback value exists; using default instead. MultiBindingExpression:target element is 'ContentControl' (Name=''); target property is 'Height' (type 'Double')

System.Windows.Data Information: 10 :
Cannot retrieve value using the binding and no valid fallback value exists; using default instead. MultiBindingExpression:target element is 'TranslateTransform' (HashCode=21021397); target property is 'Y' (type 'Double')_

I am seeing this across several of the control styles, for example in MahApps.Styles.TextBox:

```
<TranslateTransform x:Name="FloatingMessageTransform">
        <TranslateTransform.Y>
            <MultiBinding Converter="{mahConverters:MathSubtractConverter}">
                <Binding ElementName="PART_FloatingMessage"
                         Mode="OneWay"
                         Path="ActualHeight" />
                <Binding ElementName="PART_FloatingMessageContainer"
                         Mode="OneWay"
                         Path="ActualHeight" />
            </MultiBinding>
        </TranslateTransform.Y>
    </TranslateTransform>
```

As the messages are stating there aren't any fall back values defined.

When I define the value(s) then the messages are not shown (resolved) and performance is improved.

```
<TranslateTransform x:Name="FloatingMessageTransform">
        <TranslateTransform.Y>
            <MultiBinding Converter="{mahConverters:MathSubtractConverter}" FallbackValue="0">
                <Binding ElementName="PART_FloatingMessage"
                         Mode="OneWay"
                         Path="ActualHeight" />
                <Binding ElementName="PART_FloatingMessageContainer"
                         Mode="OneWay"
                         Path="ActualHeight" />
            </MultiBinding>
        </TranslateTransform.Y>
    </TranslateTransform>
```

I'm reporting this as a bug, I can create a pull requests as I find and fix the issues with the style if required.

MahApps: 2.4.9
.Net 4.7.2
WPF application.

PR #4241, opened 2022-01-02
Fix missing FallbackValue values in binding for MahApps styles

oriches:

MenuItem causing 'System.Windows.Data.Inforamtion 10:...' messages being generated. This causes performance problems when the controls are unloaded from the view or when DataContext is set to null.

XAML binding performance bug - missing FallbackValue values causes WPF information messages to be generated.

## Unit test

No unit tests - changes to XAML only

### DIFF
--- a/src/MahApps.Metro/Styles/Controls.MenuItem.xaml
+++ b/src/MahApps.Metro/Styles/Controls.MenuItem.xaml
@@ -37,12 +37,12 @@
     <Style x:Key="MahApps.Styles.MenuItem" TargetType="{x:Type MenuItem}">
         <Setter Property="Background" Value="{DynamicResource MahApps.Brushes.MenuItem.Background}" />
         <Setter Property="BorderThickness" Value="1" />
-        <Setter Property="HorizontalContentAlignment" Value="{Binding HorizontalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static HorizontalAlignment.Stretch}}" />
+        <Setter Property="HorizontalContentAlignment" Value="{Binding HorizontalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static HorizontalAlignment.Left}}" />
         <Setter Property="RenderOptions.ClearTypeHint" Value="Enabled" />
         <Setter Property="ScrollViewer.PanningMode" Value="Both" />
         <Setter Property="Stylus.IsFlicksEnabled" Value="False" />
         <Setter Property="Template" Value="{DynamicResource {ComponentResourceKey ResourceId=SubmenuItemTemplateKey, TypeInTargetAssembly={x:Type MenuItem}}}" />
-        <Setter Property="VerticalContentAlignment" Value="{Binding VerticalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static VerticalAlignment.Stretch}}" />
+        <Setter Property="VerticalContentAlignment" Value="{Binding VerticalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static VerticalAlignment.Top}}" />
         <Setter Property="mah:ControlsHelper.CornerRadius" Value="{Binding Path=(mah:ControlsHelper.CornerRadius), FallbackValue=0, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}}" />
         <Style.Triggers>
             <Trigger Property="Role" Value="TopLevelHeader">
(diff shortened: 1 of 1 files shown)


punker76:
@oriches Thx for your contribution. Can you show where or when this issue happens? If you have an example then it makes easier to understand. Thx

oriches:
Background:

I orginally notice when using MahApps.Dialogs with a fade out transition there was a slight UI freeze / judder on the fading out transition when the dialog was showing an ItemsControl (less than 10 items), similiarly I also notice when clearing a DataGrid hosted in the normal window there was longer than expected UI freeze when clearing the rows from the grid - in this example we are talking 75 columns, and less than 300 rows (financial spreadsheet kind of view)

I use a little trick with a DispatcherTimer to observe when the Dispatcher thread is overloaded (UI freezes for more than 500ms) and could see in debug mode values greater than 500ms...

Trying to understand the delay when clearing repeative templates from the view lead me down the route of inspecting the visual tree using Snoop, and from there enabling the Debug Listener for the Information level, at which point lots of repeative 'System.Windows.Data.Inforamtion 10:' messages are produced when clearing items from ItemsControls or DataGrid - this does depend on the Template one has defined obiviously, I also saw issues with ComboBox, ListView etc.

I have reported one for TextBox before, I found this one when using TextBox in the ItemTemplate for the message dialog - fixing this removed the UI stutter on the fade out transition - see `[here](https://github.com/MahApps/MahApps.Metro/issues/4217)`

I started to look for this issue in other places and found quite a few more, but for these I didn't raise a bug, I just fixed by redefining the MahApp Styles in my application. But then I thought why not contribute back (I love MahApps :)).

Regarding above issue:

On another view I have a DataGrid where when the data is refreshed everything is rebuilt - columns & rows, and each column has right-click context menu associated to it. When I enabled WPF tracing I started to see the issue with the horizontal & vertical alignments.

Now you could argue this is not a big issue when not using tabulated data but for me it is an issue becuase it slows performance and degrades the use of transitions for dialogs, flyouts etc...

Thoughts?

timunie:
Hi @oriches,

 Interesting observation. I also noticed some performance "issues" with grouping and sorting in a `DataGrid`. I wonder if this is related. 

@punker76 if I find the time in the next days I'll test this PR in my App and see if it has a noticeable benefit.

Happy coding 
Tim

oriches:
Just to be clear, this PR only affects menu items, it will affect all menu items in an app if they use the default menu style.ThanksOllieOn 3 Jan 2022 18:45, Tim ***@***.***> wrote:
Hi @oriches,
Interesting observation. I also noticed some performance "issues" with grouping and sorting in a DataGrid. I wonder if this is related.
@punker76 if I find the time in the next days I'll test this PR in my App and see if it has a noticeable benefit.
Happy coding
Tim

src/MahApps.Metro/Styles/Controls.MenuItem.xaml
```diff
@@ -37,12 +37,12 @@
     <Style x:Key="MahApps.Styles.MenuItem" TargetType="{x:Type MenuItem}">
         <Setter Property="Background" Value="{DynamicResource MahApps.Brushes.MenuItem.Background}" />
         <Setter Property="BorderThickness" Value="1" />
-        <Setter Property="HorizontalContentAlignment" Value="{Binding HorizontalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}}" />
+        <Setter Property="HorizontalContentAlignment" Value="{Binding HorizontalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static HorizontalAlignment.Stretch}}" />
```
punker76, line 40: I would set the fallback to `HorizontalAlignment.Left` because it's the default of `HorizontalContentAlignment`.

src/MahApps.Metro/Styles/Controls.MenuItem.xaml
```diff
@@ -37,12 +37,12 @@
...
         <Setter Property="RenderOptions.ClearTypeHint" Value="Enabled" />
         <Setter Property="ScrollViewer.PanningMode" Value="Both" />
         <Setter Property="Stylus.IsFlicksEnabled" Value="False" />
         <Setter Property="Template" Value="{DynamicResource {ComponentResourceKey ResourceId=SubmenuItemTemplateKey, TypeInTargetAssembly={x:Type MenuItem}}}" />
-        <Setter Property="VerticalContentAlignment" Value="{Binding VerticalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}}" />
+        <Setter Property="VerticalContentAlignment" Value="{Binding VerticalContentAlignment, RelativeSource={RelativeSource AncestorType={x:Type ItemsControl}}, FallbackValue={x:Static VerticalAlignment.Stretch}}" />
```
punker76, line 45: I would set the fallback to `VerticalAlignment.Center` because it's the default of `VerticalContentAlignment` of the MenuBase (ContextMenu).

punker76 (changes requested):

oriches:
agreed

punker76 (approved):

punker76:
@oriches your changes have been merged, thanks for your contribution 👍

PR #4300, opened 2022-05-20
Update ControlzEx to version 6.0

batzen:
This PR updates ControlzEx to version 6.0.

Because of this i also added a dependency to CSWin32 as ControlzEx does no longer expose any win32 APIs.


batzen:
@punker76 Any chance you can have a look at this?

Issue #3328, opened 2018-09-13
Some additional Windows 10 styles


Nuklon:
Like this (hope it helps):

Most changes would be a slightly thicker border / different arrow.

[x] Flat button:
![img](https://i.imgur.com/rnvksI9.gif)

[x] Radio button:

![img](https://i.imgur.com/3fwTwAS.gif)

[x] Checkbox:

![img](https://i.imgur.com/g8Jwi7J.gif)

Combobox:

![img](https://i.imgur.com/h058g2k.gif)

Textbox:

![img](https://i.imgur.com/vGN8CMf.gif)

Scrollbar:

![img](https://i.imgur.com/UwRSCNl.gif)

I might've forgotten a few, if an image/gif needs to be added, feel free to ask.

PR #4331, opened 2022-07-15
Add Win10 and Fluent TextBox styles

### LINKED ISSUE #3209: TextBox IsEnabled windows 10 style
Hey. This is not a problem - this is an offer. In windows 10 for TextBox, this style is used if it is not available:
image

Perhaps it's worth adding this feature to TextBoxHelper

punker76:
GH-3209 asks for the look Windows 10 gives a text box that is switched off, and GH-3328 collects the Windows 10 styles the library is still missing. A text box brings both now: `MahApps.Styles.TextBox.Win10` and `MahApps.Styles.TextBox.Fluent`.

The two share a template and differ in what they draw. Win10 is a fill darker than the page, a frame two pixels thick all the way round, and the box turning white with the accent around it once the caret is in. Fluent is nearly the opposite: a light translucent fill, a border you have to look for that is a touch stronger along its bottom edge, and with the caret a solid fill and a line of accent under it, corners rounded. The delete button of both comes and goes with the caret, the way the UWP box does, while a button carrying a command of the application's own stays where it is.

What each of them is made of sits in the theme, `MahApps.Brushes.TextControl` for the one and `MahApps.Brushes.TextControl.Fluent` for the other, the Fluent set taken from the WinUI values. The thicknesses and the padding are resources as well, so a style of your own can answer them differently without a template of its own.

`ControlsHelper` grew `DisabledBorderBrush` for the border of a control with nothing left to say, which is what GH-3209 is about, and both styles point it at what their own set has for that state.

One thing fell out of this that is not about the two new styles. The content of a text box sat beside its border rather than inside it, so with a corner radius everything square went on past the round end: the background the delete button paints under the mouse, the white a focused Win10 box turns, the highlight of a selection. The content is the child of the border now and is clipped to the corners the border leaves over, the delete button takes the two corners of the side it stands on, and a chromeless button clips its content the same way, since that is the style the button in a text box wears. `MahApps.Styles.Button.TextControl.Delete` stands in the buttons file with the rest of them, where an application can put its own in front of it.

The demo has a page for it. Every style the library has for a text control side by side, empty, with text and switched off, and one card per subject for what the attached properties of `TextBoxHelper`, `PasswordBoxHelper` and `ControlsHelper` do to one. Everything that lifts its watermark out of the box stands in a row of its own, because that lift grows the control it belongs to and moves whatever is under it.

BREAKING CHANGE: the background of the dark theme changes from `#FF252525` to `#FF141414`. That is a change for every theme and every control, and the two new styles do not depend on it: the fill of a Win10 box is forty percent black over whatever the page is, so the box keeps its distance from the page either way.

Closes #3209

pr-1-seite-dunkel

pr-2-seite-hell

pr-3-styles

pr-4-floating

pr-5-fokus

### DIFF
--- a/src/MahApps.Metro/Controls/Helper/ControlsHelper.cs
+++ b/src/MahApps.Metro/Controls/Helper/ControlsHelper.cs
@@ -223,6 +223,43 @@ namespace MahApps.Metro.Controls
             obj.SetValue(MouseOverBorderBrushProperty, value);
         }
 
+        public static readonly DependencyProperty DisabledBorderBrushProperty
+            = DependencyProperty.RegisterAttached(
+                "DisabledBorderBrush",
+                typeof(Brush),
+                typeof(ControlsHelper),
+                new FrameworkPropertyMetadata(Brushes.Transparent, FrameworkPropertyMetadataOptions.AffectsRender | FrameworkPropertyMetadataOptions.Inherits));
+
+        /// <summary>
+        /// Gets the brush used to draw the disabled border brush.
+        /// </summary>
+        [Category(AppName.MahApps)]
+        [AttachedPropertyBrowsableForType(typeof(TextBox))]
+        [AttachedPropertyBrowsableForType(typeof(CheckBox))]
+        [AttachedPropertyBrowsableForType(typeof(RadioButton))]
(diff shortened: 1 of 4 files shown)


JackDjTom:
Any idea when this issue will be fixed and version 3.0 will be released approximately?

PR #4354, opened 2022-10-03
Adding the ability for the dialog title to contain arbitrary content

### LINKED ISSUE #4350: Allow arbitrary content for metro dialog.
## Describe the feature

Right now (version 2.4.9) the `BaseMetroDialog` has a `Title` property that is of type string. I would prefer if the dialog allowed arbitrary content to allow for putting more (in my case an icon plus some text). For now my work around is to skip using that property and instead use my own template for it.

## Additional context

I am happy to submit a PR for this if this is a desirable change.
I took a look through the code and there are a few key changes that warrant some discussion.

1. The `MetroDialogAutomationPeer` currently uses the `Title` property as one possible location where it can get a return value for `GetNameCore()`. This could be changed to only use this if the `Title` property contains a string (matching existing behavior), or it could do something like call `.ToString()` on the `Title` property (possibly giving back an ugly looking type name).
2. There are several wrapper methods that expose the title as a string (such as `ProgressDialogController.SetTitle`, and many methods on `DialogManager`). These could either be replaced to accept `object` rather than `string` as the parameter type (a binary braking change), or they could be overloaded (potentially accidently invoking the new overload if someone was previously passing a type that has an implicit string conversion).

Keboo:
Fixes #4350

### DIFF
--- a/src/MahApps.Metro.Samples/MahApps.Metro.Demo/MainWindow.xaml
+++ b/src/MahApps.Metro.Samples/MahApps.Metro.Demo/MainWindow.xaml
@@ -1,13 +1,13 @@
 ﻿<mah:MetroWindow x:Class="MetroDemo.MainWindow"
                  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
                  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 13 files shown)


punker76:
@Keboo Thank you ❤️ your changes have been merged, thanks for your contribution 👍

PR #4407, opened 2023-06-08
when restoring window size use the DPI scale factor

ykarpeev:

I noticed that restoring the window size wasn't very consistent - and on a brand new test project it would get smaller each time it was opened and closed. Reviewing the code it looks like WindowPlacementSetting uses the DPI to adjust the scale factor when saving the settings - but that isn't used when restoring the window.  This change takes the DPI scale factor into account when restoring. 

This fixes the issue for me. Don't have enough experience with the codebase to add a unit test.

### DIFF
--- a/src/MahApps.Metro/Controls/WinApiHelper.cs
+++ b/src/MahApps.Metro/Controls/WinApiHelper.cs
@@ -89,10 +89,13 @@ namespace MahApps.Metro.Controls
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


ykarpeev:
@dotnet-policy-service agree

punker76:
@ykarpeev your changes have been merged, thanks for your contribution 👍

PR #4451, opened 2023-12-21
Add an option to always show ErrorTemplate text

nekizalb:
Add an option to always show ErrorTemplate text, regardless of focus or mouse over.

## This change extends the Validation Error Template text show options to include an 'always show' option. 

The existing options of showing on focus or on mouseover did not work well for me when trying to apply the error template to a border surrounding a group of radio buttons used primarily on a touch interface. Mouseover isn't a thing on touch, and focus meant the text would only show when a radio button was selected, which removed the error entirely. 

This change introduces a new attached property in ValidationHelper to enable the behavior and a new style trigger that will enable when the template has errors and the new attached property, ValidationHelper.AlwaysShowValidationError, is set to true for the element. 

## Unit test

I did not find existing unit testing of the ValidationHelper properties to model any new test after.

### DIFF
--- a/src/MahApps.Metro/Controls/Helper/ValidationHelper.cs
+++ b/src/MahApps.Metro/Controls/Helper/ValidationHelper.cs
@@ -67,5 +67,35 @@ namespace MahApps.Metro.Controls
         {
             element.SetValue(ShowValidationErrorOnMouseOverProperty, BooleanBoxes.Box(value));
         }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


tyler-lee-ati:
Apologies for the delay; with the holidays, it took me a bit to get confirmation

@dotnet-policy-service agree company="Aristocrat Technologies"

nekizalb:
Apologies for the delay; holidays prevented me getting confirmation

@dotnet-policy-service agree company="Aristocrat Technologies"

PR #4478, opened 2024-04-10
Make MultiSelectionComboBox DropDown respect MaxDropDownHeight value

Ifry:
Currently MultiSelectionComboBox drop down height is always 200 and can't be changed. It should respect MaxDropDownHeight value. 
Change would make behavior similar to ComboBox style


Ifry:
@dotnet-policy-service agree

timunie:
lgtm, however I can't merge PRs here. So ping @punker76 feel free to merge when you're also okay with that improvement.

punker76:
@Ifry your changes have been merged, thanks for your contribution 👍

PR #4508, opened 2024-07-24
#2166 automatiopeer for numericupdown

### LINKED ISSUE #2166: Controls Do not get distinct automation ID's

noubar:
I have added automationpeer to numericupdown ui element.
The automation peer will let the numericupdown button to get automationids and name given from user which will ease the automation testing

Closes #2166


noubar:
@dotnet-policy-service agree

PR #4527, opened 2024-10-30
feat: Add ChangeFocusWithEnterOnUsernameField to LoginDialogSettings

Valmir-Cunha:
- Implemented `ChangeFocusWithEnterOnUsernameField` property to manage focus behavior when the Enter key is pressed in the username field.
- Added corresponding dependency property registration and getter/setter methods.
- Updated key handling logic to support the new property.


Valmir-Cunha:
@dotnet-policy-service agree company="ZeusDev"

Issue #4551, opened 2025-06-09
DateTimePicker returns mixed DateTimeKind values  (Unspecified vs Local)


XingbinChen:
Hi,
I’m using the DateTimePicker control and binding its SelectedDateTime property to a DateTime? property in my ViewModel.

## Bug
If I pick the date first, the SelectedDateTime.Kind is Unspecified.

If I pick the time first, the Kind becomes Local.

## Why it happens
Looking at the source:

CalendarSelectedDateChanged (inside the DateTimePicker.calendar part) returns a DateTime with Kind == Unspecified.

ClockSelectedTimeChanged (inside TimePickerBase) sets SelectedDateTime = DateTime.Today when the previous value is null; DateTime.Today has Kind == Local.

## Why this is a problem
Consumers of the control have to handle two different DateTimeKind cases, which is easy to overlook and may introduce bugs.

## Possible fix
Replace DateTime.Today with DateTimeOffset.Now.Date (or another DateTime that keeps Kind == Unspecified) so that SelectedDateTime.Kind is consistently Unspecified, no matter which part (date or time) the user selects first.

## Environment

```
MahApps.Metro version: v2.4.10
Target Framework: .Net 8
```

## Screenshots

![Image](https://github.com/user-attachments/assets/3f192c80-5a1b-48db-89df-d32af1e494bd)

PR #4565, opened 2026-04-14
Fix 7 critical bugs in NumericUpDown and DateTimePicker

Thuram2003:
This PR fixes multiple critical issues in NumericUpDown and DateTimePicker controls in MahApps.Metro, improving stability, usability, and performance.

🔧 Key Fixes
Fixed keyboard input blocking in NumericUpDown
Fixed inconsistent DateTimeKind in DateTimePicker
Prevented duplicate ValueChanged event firing
Replaced unsafe async void method to improve stability
Fixed potential integer overflow in hexadecimal formatting
Improved event handler cleanup to prevent memory leaks
Removed unnecessary double value coercion for better performance

🎯 Impact
Better user input experience (smoother typing)
More consistent and reliable date handling
Improved performance and memory safety
Reduced risk of crashes and unexpected behavior

🧪 Testing
Verified compilation across target frameworks (.NET Framework & .NET 6/8)
Manually tested all affected controls
No breaking changes introduced
📌 Notes
All changes are backward compatible and focused on improving internal control behavior without affecting public APIs.

### DIFF
--- a/src/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/src/MahApps.Metro/Controls/NumericUpDown.cs
@@ -1020,6 +1020,19 @@ namespace MahApps.Metro.Controls
         {
             base.OnApplyTemplate();
 
+            // Clean up old event handlers to prevent memory leaks
+            if (this.repeatUp != null)
+            {
+                this.repeatUp.Click -= this.OnRepeatUpClick;
+                this.repeatUp.PreviewMouseUp -= this.OnRepeatButtonMouseUp;
+            }
+
+            if (this.repeatDown != null)
+            {
+                this.repeatDown.Click -= this.OnRepeatDownClick;
+                this.repeatDown.PreviewMouseUp -= this.OnRepeatButtonMouseUp;
+            }
+
             this.repeatUp = this.GetTemplateChild(PART_NumericUp) as RepeatButton;
             this.repeatDown = this.GetTemplateChild(PART_NumericDown) as RepeatButton;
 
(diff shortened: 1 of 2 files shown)


Thuram2003:
@dotnet-policy-service agree

punker76:
Thanks for taking the time on this. I went through the seven changes one by one and wrote a test for each claim first, so that every defect is reproduced before the fix is applied. Three of them hold up, two are behaviour-neutral, and two I cannot take as they are. Details below, since the reasoning matters more than the verdict.

## Confirmed, taken over with tests

**Duplicate ValueChanged.** `ChangeValueFromTextInput` called `OnValueChanged` on top of the call the `ValueProperty` callback had already made. Typing `42` reports six transitions where three are expected:

```
0 ->  | 0 ->  |  -> 4 |  -> 4 | 4 -> 42 | 4 -> 42
```

The count alone is not the contract, because typing walks the value through several states. The test asserts that no single transition is reported twice.

**Swallowed keystrokes below Minimum.** With `Minimum` at 50, the `5` of `55` never arrived: `OnPreviewTextInput` ran the coercion check against the intermediate text and set `e.Handled`. The test raises `PreviewTextInput` and asserts on `Handled`, which fails before the change.

**Accumulating repeat button handlers.** Your point about named methods is right, and the effect is observable: after a second template pass a single click changes the value twice, because the lambdas of the first pass are still attached to the same button. Test asserts `Value == 1` after one click and gets `2` without the fix.

## Behaviour-neutral

The `async void` cleanup is correct. `await Task.FromResult(true)` on an already-completed task runs synchronously, so removing it changes nothing observable.

Removing the manual coercion in `SetValueTo` is also fine, since `ValueProperty` is registered with a `CoerceValueCallback` that does the same work, and `SetValueTo` already clamps before it. One subtlety worth knowing: the base value stored in the property differs afterwards, so a later change of `Minimum` or `Maximum` re-coerces from a different starting point.

Neither has an observable effect that a test can pin, so both would be better placed in a separate refactoring commit than in a bug-fix PR.

## Cannot take as written

**The hexadecimal guard would crash instead of misformat.** Returning `false` for values outside the int range sends them to the non-hex branch of `FormattedValueString`, which ends at `newValue.ToString(format, culture)`. For format `X` that is not a valid numeric format string for `double`, so it throws a `FormatException`. Verified:

| value | current | with this PR |
| --- | --- | --- |
| `3000000000` | `7FFFFFFF` | `FormatException` |
| `2.7` | `2` | `FormatException` |

The overflow you spotted is real, and `7FFFFFFF` for three billion is wrong. But the fix has to keep the caller supplied with a formattable result, so `return false` is not the place to stop. Same for the new decimal check: `(int)2.7` currently renders as `2`, which is at least a number.

**The `AssumeLocal` change does not fix #4551, it deepens it.** With it, typed input stores `DateTimeKind.Local` while `CalendarSelectedDateChanged` keeps producing `Unspecified`. That issue asks for one consistent kind, and this leaves two. The actual origin is the `DateTime.Today` fallback in `TimePickerBase.ClockSelectedTimeChanged`, which is the one path that introduces `Local`. Making that fallback kind-less brings all three paths in line, and that is the route I took.

punker76:
A short follow-up, because your overflow finding did lead to a fix.

The saturation is gone, though at a different spot than this PR put it. The operand now stays an `int` inside the int range, so negative values keep the 32 bit form that existing test cases pin (`-1` as `ffffffff`), and becomes a `long` outside it. Three billion renders as `B2D05E00` instead of `7FFFFFFF`, and `int.MaxValue + 1` as `80000000`. Beyond the long range it still saturates, since no integral type is left for that. The formatter never bails out, so the caller always receives something it can display.

On how this goes forward: I am taking the three confirmed fixes, but I will commit them through a separate PR of my own rather than merging this one. That PR adds the tests that reproduce each of the three defects before the change is applied, and it splits the two controls, because the `DateTimePicker` side needs the #4551 discussion and the `NumericUpDown` side does not. The commits reference this PR for where the analysis came from.

So there is nothing left for you to do here. Thanks for the groundwork, in particular the overflow and the accumulating handlers, which were both real.

punker76:
@Thuram2003 part of your changes have been merged, thanks for your contribution 👍

PR #4567, opened 2026-06-07
fix: add JSON fallback for SaveWindowPosition on .NET 9

### LINKED ISSUE #4541: .NET 9 breaks SaveWindowPosition
## Describe the bug
If `SaveWindowPosition = true` in the MetroWindow and `TargetFramework = net9.0-windows` the application throws an exception on closing and not saving the window position:

```
ZH.TED.Test.WPF Error: 0 : MahApps.Metro.Behaviors.WindowsSettingBehavior: The settings could not be saved! System.Configuration.ConfigurationErrorsException: Failed to save settings: An error occurred loading a configuration file: An unexpected error occurred in 'ClientConfigurationHost::OpenStreamForWrite '' '''.
 ---> System.Configuration.ConfigurationErrorsException: An error occurred loading a configuration file: An unexpected error occurred in 'ClientConfigurationHost::OpenStreamForWrite '' '''.
 ---> System.InvalidOperationException: An unexpected error occurred in 'ClientConfigurationHost::OpenStreamForWrite '' '''.
   at System.Configuration.ClientConfigurationHost.OpenStreamForWrite(String streamName, String templateStreamName, Object& writeContext)
   at System.Configuration.MgmtConfigurationRecord.SaveAs(String filename, ConfigurationSaveMode saveMode, Boolean forceUpdateAll)
   --- End of inner exception stack trace ---
   at System.Configuration.MgmtConfigurationRecord.SaveAs(String filename, ConfigurationSaveMode saveMode, Boolean forceUpdateAll)
   at System.Configuration.ClientSettingsStore.WriteSettings(String sectionName, Boolean isRoaming, IDictionary newSettings)
   --- End of inner exception stack trace ---
   at System.Configuration.ClientSettingsStore.WriteSettings(String sectionName, Boolean isRoaming, IDictionary newSettings)
   at System.Configuration.LocalFileSettingsProvider.SetPropertyValues(SettingsContext context, SettingsPropertyValueCollection values)
   at System.Configuration.SettingsBase.SaveCore()
   at System.Configuration.SettingsBase.Save()
   at MahApps.Metro.Behaviors.WindowsSettingBehavior.SaveWindowState()
```

## Steps to reproduce
Here is an example project:
[ZH.TED.Test.WPF.zip](https://github.com/user-attachments/files/18586450/ZH.TED.Test.WPF.zip)

If you switch to net8.0-windows as TargetFramework, it works as expected.

## Environment

```
MahApps.Metro version: 2.4.10
Target Framework: .Net 9 (9.0.100)
```

JyslaFancy:
- ApplicationSettingsBase.Save() fails on .NET 9 < 9.0.3 due to a runtime bug (dotnet/runtime#112410) where ClientConfigurationHost receives an empty path on UNC/network locations.
- SaveWindowState() now saves a JSON backup to %LocalAppData% as a fallback when the settings save fails.
- LoadWindowState() tries the JSON fallback when settings reload fails or returns an empty placement.
- On successful save via ApplicationSettingsBase, the JSON backup is also updated to stay in sync.

## Description

Fixes #4541 — .NET 9 breaks SaveWindowPosition in MetroWindow.

### Problem
On .NET 9 prior to version 9.0.3, ApplicationSettingsBase.Save() throws a 
ConfigurationErrorsException when the project path is on a UNC/network location. 
This is a known .NET runtime bug (dotnet/runtime#112410), fixed in .NET 9.0.3.

### Fix
Adds a JSON-based fallback for window placement persistence:
- Save: if settings.Save() fails → placement saved to 
  %LocalAppData%/<app>/WindowPlacement/<WindowType>.json
- Load: if settings.Reload() fails or returns empty → loads from JSON fallback
- On success: JSON stays in sync

No new dependencies — uses System.Text.Json.

### DIFF
--- a/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
+++ b/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
@@ -135,12 +135,14 @@ namespace MahApps.Metro.Behaviors
             }
 
             // check for existing placement and prevent empty bounds
-            if (settings.Placement is null || settings.Placement.normalPosition.IsEmpty)
+            var placement = settings.Placement;
+            if (placement is null || placement.normalPosition.IsEmpty)
             {
 #if !NET462
                 // Fallback: try to load from JSON backup if settings has no valid placement
                 if (TryLoadFromJsonFallback(window, out var fallbackPlacement))
                 {
+                    placement = fallbackPlacement;
                     settings.Placement = fallbackPlacement;
                 }
                 else
@@ -154,7 +156,7 @@ namespace MahApps.Metro.Behaviors
 
             try
             {
-                var wp = settings.Placement.ToWINDOWPLACEMENT();
+                var wp = placement.ToWINDOWPLACEMENT();
                 WinApiHelper.SetWindowPlacement(window, wp);
             }
             catch (Exception ex)
--- a/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
+++ b/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
@@ -143,13 +143,13 @@ namespace MahApps.Metro.Behaviors
                 {
                     settings.Placement = fallbackPlacement;
                 }
-#endif
-            }
-
-            // If we still have no valid placement, nothing to restore
-            if (settings.Placement is null || settings.Placement.normalPosition.IsEmpty)
-            {
+                else
+                {
+                    return;
+                }
+#else
                 return;
+#endif
             }
 
             try
--- a/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
+++ b/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
@@ -156,7 +156,7 @@ namespace MahApps.Metro.Behaviors
 
             try
             {
-                var wp = placement.ToWINDOWPLACEMENT();
+                var wp = placement!.ToWINDOWPLACEMENT();
                 WinApiHelper.SetWindowPlacement(window, wp);
             }
             catch (Exception ex)
--- a/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
+++ b/src/MahApps.Metro/Behaviors/WindowsSettingBehavior.cs
@@ -4,7 +4,9 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Runtime.InteropServices;
+using System.Text.Json;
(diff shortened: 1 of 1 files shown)


punker76:
Thanks for digging into this, and for tracking it down to dotnet/runtime#112410. The diagnosis is right and the problem is real.

Before this goes in, one finding from testing the branch, because it changes behaviour in a way that is easy to miss.

## Loading throws now when the settings are broken

`LoadWindowState` drops the `return` in the `catch`, but the line after it reads `settings.Placement`, and that is outside the `try`. The default store throws `MahAppsException` from that getter when the configuration file is corrupted, so the exception now travels up through `OnAttached` into `Window.Show()`.

I ran the same test against both versions, with a settings store whose `Reload` and `Placement` both throw, which is what a corrupted file looks like:

| | `window.Show()` |
| --- | --- |
| develop | returns, window opens |
| this branch | throws `MahAppsException: the settings file seems to be corrupted` |

A lost window position turns into a window that will not open, for the exact case the current code deliberately swallows.

## On the approach

The rest is a judgement call rather than a defect, but worth saying:

- it adds a second store to the library permanently, for a runtime bug that Microsoft fixed in 9.0.3
- it writes into `%LocalAppData%` for every application that saves its window position, with no way to turn that off
- the fallback catches every save failure, not just this one, so missing permissions or a broken file are masked from then on
- both stores are written on success and either can win on load, so a stale file can outlive a deliberate reset
- behaviour differs per target framework through `#if !NET462`

## What we went with instead

`MetroWindow.WindowPlacementSettings` already takes any `IWindowPlacementSettings`, and `GetWindowPlacementSettings()` is virtual, so the storage was always the caller's to choose. That path works today, which we checked rather than assumed: the behavior saves and loads through a store supplied by the caller.

So rather than making every application write two files, `develop` now carries `WindowPlacementFileSettings`, a store that writes a file and is used only when asked for:

```csharp
this.WindowPlacementSettings = WindowPlacementFileSettings.ForWindow(this);
```

Your code was a good starting point for it. It writes plain numbers through `XmlSerializer` instead of the WPF types, so it behaves the same on net462 too, and the `MetroWindow` documentation page now explains the choice.

Given that, we would rather not take an automatic fallback into the library itself. Thank you again for the report and the analysis, tracking it to the runtime issue is what made the opt-in store possible. If you see a case it does not cover, please say so.

punker76:
Closing in favour of #4597, which adds `WindowPlacementFileSettings` as an opt-in store rather than an automatic fallback. Your report and the analysis are what made that possible, thank you.

PR #4568, opened 2026-06-12
fix(MetroWindow): add GlowBrush dependency property (fixes #4564)

### LINKED ISSUE #4564: GlowBrush invalid in MetroWindow
## Describe the bug
GlowBrush invalid property on MetroWindow

## Steps to reproduce

Follow example on docs site. Create new window, add controls tag and change window type.
Add Glowbrush as described in docs and it is not valid in MetroWindow

JyslaFancy:
Fixes #4564 — GlowBrush invalid property on MetroWindow.

### Problem
The MahApps.Metro documentation references a GlowBrush property on MetroWindow, but this property was never implemented. Users following the docs example get 'GlowBrush invalid property' at design time.

### Root cause
MetroWindow inherits GlowColor (Color?) from ControlzEx.WindowChromeWindow, but no GlowBrush convenience property was exposed.

### Fix
Added a GlowBrush dependency property (of type Brush) that wraps the inherited GlowColor:
- Setting GlowBrush to a SolidColorBrush extracts the Color and assigns it to GlowColor
- Setting GlowBrush to null clears GlowColor
- The getter returns the current GlowBrushProperty value

This matches the pattern used by other brush properties on MetroWindow (e.g. NonActiveBorderBrush).

### DIFF
--- a/src/MahApps.Metro/Controls/MetroWindow.cs
+++ b/src/MahApps.Metro/Controls/MetroWindow.cs
@@ -558,6 +558,35 @@ namespace MahApps.Metro.Controls
             set => this.SetValue(NonActiveBorderBrushProperty, value);
         }
 
+        /// <summary>Identifies the <see cref="GlowBrush"/> dependency property.</summary>
+        public static readonly DependencyProperty GlowBrushProperty
+            = DependencyProperty.Register(nameof(GlowBrush),
+                                          typeof(Brush),
+                                          typeof(MetroWindow),
+                                          new PropertyMetadata(default(Brush), OnGlowBrushChanged));
+
+        /// <summary>
+        /// Gets or sets the brush used for the glow effect of the window border when the window is active.
+        /// This property wraps the inherited <see cref="GlowColor"/> property from ControlzEx.
+        /// </summary>
+        public Brush GlowBrush
+        {
+            get => (Brush)this.GetValue(GlowBrushProperty);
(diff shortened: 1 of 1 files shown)


BinToss:
I believe this should be changed in ControlzEx, instead. The GlowBrush/GlowColor issue originated from ControlzEx, after all—¹not MahApps.Metro.

```cs
        /// <inheritdoc cref="GlowWindowBehavior.GlowColor"/>
        [Obsolete($"Use {nameof(GlowColor)}")]
        public Color? GlowBrush
        {
            get => this.GlowColor;
            set => this.GlowColor = value;
        }
```
Or throw `NotImplementedException`. I'd prefer that it cause build errors instead of build warnings and runtime errors, but I don't know how to do that. Maybe a custom analyzer that checks for invalid references to `GlowBrush` and suggests using refactoring to `GlowColor`.

P.S. The removal of `GlowBrush` is obviously a breaking change, but breaking changes in transitive dependencies are often overlooked when upgrading explicit dependencies. It's no wonder this became an issue.

¹key-combo'ed em-dash and superscript. They're not LLM-generated.

punker76:
`GlowBrush` and `NonActiveGlowBrush` was removed in MahApps with ControlzEx v6. You should use `GlowColor` and `NonActiveGlowColor` now (pre-release v3.0.0 of MahApps).

PR #4607, opened 2026-09-08
fix(NumericUpDown): show a value the way it would be typed

### LINKED ISSUE #3673: NumericUpDown Exponent displayed for decimal type

punker76:
Closes #3673

Without a `StringFormat` the value went straight to `double.ToString()`, which answers `5E-05` for a small number. Since .NET Core 3.0 it also writes every digit a calculation left behind, so `0.1 + 0.2` reached the text box as `0.30000000000000004`. Neither belongs in a control nobody asked to format anything.

| value | before | now |
| --- | --- | --- |
| `0.00005` | `5E-05` | `0.00005` |
| `0.1 + 0.2` | `0.30000000000000004` | `0.3` |
| `1d / 3d` | `0.3333333333333333` | `0.333333333333333` |
| `1e-7` | `1E-07` | `0.0000001` |
| `1.23e-15` | `1.23E-15` | `0.00000000000000123` |
| `2.5`, `0.1`, `123.456`, `-42.75` | unchanged | unchanged |
| `1234567890123456` | unchanged | unchanged |
| `1e20`, `1e-16`, `double.MaxValue` | unchanged | unchanged |

### Where the bounds come from

Plain decimal notation is used while the magnitude stays between `1e-15` and `1e15`.

Past the upper one a decimal format would drop digits: a custom format rounds at the fifteenth significant digit, so `1234567890123456` would come out as `1234567890123460`. Past the lower one the zeroes in front of the number take over, `1e-20` reading as `0.00000000000000000001`. In both of those the exponent is the better answer, so outside the bounds the framework still decides, exactly as it did everywhere before.

### What this changes for anyone using it

An application that sets no `StringFormat` will show calculated values differently, which is the point of the change. A `StringFormat` of your own decides everything as before, and a test pins that down.

The sixteenth and seventeenth significant digit are no longer shown, though they are still held in the value. Only code that pushes `double` to its limit would notice.

### Not in here

The report also asks why a `decimal` binding loses precision, and that is not something a display format can answer: `Value` is a `double?`, so a `decimal` is converted on the way in. Binding a `decimal?` also writes a `SystemConvertConverter` error to the trace whenever the value is null, which still happens. Both belong to the type, not to the formatting, and are for the typed controls that come next.

### DIFF
--- a/src/MahApps.Metro/Controls/NumericUpDown.cs
+++ b/src/MahApps.Metro/Controls/NumericUpDown.cs
@@ -1363,7 +1363,34 @@ namespace MahApps.Metro.Controls
                 }
             }
 
-            return newValue.ToString(culture);
+            return PlainValueString(newValue, culture);
+        }
+
+        /// <summary>
+        /// The lowest and the highest magnitude that reads better without an exponent. Past the upper
+        /// one a decimal format would drop digits, since a custom format rounds at the fifteenth
+        /// significant one; past the lower one the zeroes in front of the number take over.
+        /// </summary>
+        private const double SmallestPlainValue = 1e-15;
+
+        private const double LargestPlainValue = 1e15;
+
+        /// <summary>
+        /// Writes a value the way it would be typed, for as long as that reads better than an exponent.
+        /// Without this the framework decides, and it answers 5E-05 for a small number and every digit
+        /// a calculation left behind for the rest, neither of which belongs in a text box that was
+        /// never asked to format anything.
+        /// </summary>
+        private static string PlainValueString(double value, CultureInfo culture)
+        {
+            var magnitude = Math.Abs(value);
+
+            if (magnitude != 0 && (magnitude < SmallestPlainValue || magnitude >= LargestPlainValue))
+            {
+                return value.ToString(culture);
+            }
+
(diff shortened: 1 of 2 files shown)


punker76:
Closing this one. The typed controls make a different answer possible: whoever needs every digit to survive takes a DecimalUpDown, rather than a double being formatted into looking right. The display format is coming back in that branch, shaped by what the type can actually hold.

PR #4623, opened 2026-09-14
fix(FlipView): let an item out of a source name the banner

### LINKED ISSUE #4447: FlipView Banner only showing default value when using DataTemplate
## Describe the bug

The `FlipView` control in MahApps.Metro is ignoring the `BannerText` property of the `FlipViewItem`. Regardless of whether `BannerText` is set to an empty string, a non-empty string, or bound to a property, the text "Banner" is always displayed.

## Steps to reproduce

1. Create a new WPF project.
2. Install the MahApps.Metro NuGet package.
3. Add a `FlipView` control to the main window.
4. Set the `ItemTemplate` of the `FlipView` to include a `FlipViewItem` with `BannerText=""`, `BannerText="SomeText"`, or `BannerText="{Binding SomeProperty}"`.
5. Run the project and observe that the "Banner" text is displayed, regardless of the `BannerText` setting.

The code in the `ItemTemplate` is as follows:

```xml
<mah:FlipView.ItemTemplate>
    <DataTemplate>
        <mah:FlipViewItem
            BannerText=""
            BorderThickness="0"
            Foreground="White">
            <Image
                HorizontalAlignment="Stretch"
                VerticalAlignment="Stretch"
                Source="{Binding ImageSource}"
                Stretch="Fill" />
        </mah:FlipViewItem>
    </DataTemplate>
</mah:FlipView.ItemTemplate>
```

## Expected behavior

The `FlipView` control should respect the `BannerText` property of the `FlipViewItem`. The text displayed should match the value of `BannerText`.

## Actual behavior

Regardless of the `BannerText` setting, the text "Banner" is always displayed.

## Mitigation

Set IsBannerEnabled to false in FlipView and do it myself if i need it

## Environment

```
MahApps.Metro version: 3.0.0-alpha0457
Windows build number: Win10 22H2 
Visual Studio: 2022 17.8.3
Target Framework: .Net 8.0-windows
```

## Screenshots

grafik

punker76:
A `FlipView` ignored the `BannerText` of a `FlipViewItem` and kept showing the word the property starts out with, as soon as the items came from an `ItemsSource`. #4447

The control is an `ItemsControl` whose template has no `ItemsPresenter`. It shows `SelectedItem` in a `TransitioningContentControl` with `ContentTemplate="{TemplateBinding ItemTemplate}"`, so no `FlipViewItem` containers are ever made, the item container generator sits at `NotStarted`, and an `ItemContainerStyle` reaches nothing. A `FlipViewItem` inside the item template, the way the report writes it, is therefore not the wrong way round. It is the only place an item out of a source can name a banner.

It could not name one, because the item looked for the control it belongs to the way a container does, through `ItemsControlFromItemContainer`, and content is not a container. Looking up the tree finds it.

That alone still put the wrong text on the banner. Logging what the banner was told during a flip:

```
Banner to first    the item appears and the binding takes
first to second    the same item, its data context moves on
Banner to first    a second item, the presenter holds both during the fade
first to Banner    the outgoing one loses its data context
```

The last line is the damage. The item on its way out falls back to the text a `FlipViewItem` starts out with and writes that over the banner of the item that has just arrived. Only the item on show has its say now, which is the one that either is the selected item or carries it.

Closes #4447


JKamsker:
Didnt know i created that issue - but awesome to see it fixed now - thank you! <3

